PR #5, opened 2022-08-03
Add the command line tool kwokctl

wzshiming:
# Feature

``` console
Kwokctl is a Kwok cluster management tool

Usage:
  kwokctl [command] [flags]
  kwokctl [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  create      Creates one of [cluster]
  delete      Deletes one of [cluster]
  get         Gets one of [artifacts, clusters, kubeconfig]
  help        Help about any command
  kubectl     kubectl in cluster
  logs        Logs one of [etcd, kube-apiserver, kube-controller-manager, kube-scheduler, kwok-controller, prometheus]

Flags:
  -h, --help          help for kwokctl
      --name string   cluster name (default "kwok")

```

## Runtime and OS support

|         | binary | docker | nerdctl | kind |
| ------- | ------ | ------ | ------- | ---- |
| Linux   | ✔️      | ✔️      | ✔️       | ✔️    |
| Darwin  | ➕      | ✔️      | ✖️       | ✔️    |
| Windows | ➕      | ✔️      | ✖️       | ✔️    |

✔️: Support
✖️: Not support
➕: You need to use your own build of the kube binary

Not tested on Windows yet.

## Default Runtime

Linux: binary
Darwin/Windows: docker

## All-in-one image

``` bash
./images/cluster/build.sh --tag cluster:v0.1.0-kube-1.24 --kube-version 1.24.2
docker run --rm -it -p 8080:8080 cluster:v0.1.0-kube-1.24
```


wzshiming:
/hold
Waiting for review

wzshiming:
The `test/kwokctl/kwokctl.test.sh` for testing the features in this commit

pkg/kwokctl/cmd/kwokctl/get/images/images.go
```diff
@@ -0,0 +1,72 @@
...
+	Runtime string
+}
+
+// NewCommand returns a new cobra.Command for getting the list of clusters
+func NewCommand(logger logger.Logger) *cobra.Command {
+	flags := &flagpole{}
+	cmd := &cobra.Command{
+		Args:  cobra.NoArgs,
+		Use:   "images",
+		Short: "Lists images used by fake cluster, only for docker/nerdctl/kind runtime",
```
yibozhuang, line 40: same here, get the list of supported runtimes from the runtime pkg

pkg/kwokctl/cmd/kwokctl/get/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = dc.KubectlInCluster(ctx, utils.IOStreams{
+		Out:    os.Stdout,
+		ErrOut: os.Stderr,
+	}, "config", "view", "--minify", "--flatten", "--raw")
+
+	if err != nil {
```
yibozhuang, line 65: nit: this can be simplified to:
```go
return err
```

same for the other places

pkg/kwokctl/pki/openssl.cnf
```diff
@@ -0,0 +1,14 @@
...
+[ v3_req ]
+basicConstraints = CA:FALSE
+keyUsage = nonRepudiation, digitalSignature, keyEncipherment
+subjectAltName = @alt_names
+[alt_names]
+DNS.1 = kubernetes
+DNS.2 = kubernetes.default
+DNS.3 = kubernetes.default.svc
+DNS.4 = kubernetes.default.svc.cluster.local
+IP.1 = 127.0.0.1
```
yibozhuang, line 14: nit: new line

pkg/kwokctl/pki/admin.crt
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
```
yibozhuang, line 1: do we need to check in `admin.crt`, `admin.key`, and `ca.crt` if they can be generated by `DumpPki()` in `pki.go`?

pkg/kwokctl/runtime/binary/cluster.go
```diff
@@ -0,0 +1,530 @@
...
+
+	bin := utils.PathJoin(conf.Workdir, "bin")
+	kubeApiserverPath := utils.PathJoin(bin, "kube-apiserver"+vars.BinSuffix)
+	kubeControllerManagerPath := utils.PathJoin(bin, "kube-controller-manager"+vars.BinSuffix)
+	kubeSchedulerPath := utils.PathJoin(bin, "kube-scheduler"+vars.BinSuffix)
+	kwokControllerPath := utils.PathJoin(bin, "kwok-controller"+vars.BinSuffix)
+	etcdPath := utils.PathJoin(bin, "etcd")
+	prometheusPath := utils.PathJoin(bin, "prometheus")
+
+	err = utils.ForkExecKill(ctx, conf.Workdir, kwokControllerPath)
```
yibozhuang, line 424: for later: maybe we can think about doing the `ForkExecKill` of each component in parallel rather than serial (same for bring up)

pkg/kwokctl/runtime/binary/prometheus.yaml.tpl
```diff
@@ -0,0 +1,111 @@
...
+- job_name: "kube-scheduler"
+  scheme: http
+  honor_timestamps: true
+  metrics_path: /metrics
+  follow_redirects: true
+  enable_http2: true
+  static_configs:
+  - targets:
+    - localhost:{{ .KubeSchedulerPort }}
+{{ end }}
```
yibozhuang, line 111: new line.

pkg/kwokctl/runtime/compose/prometheus.yaml.tpl
```diff
@@ -0,0 +1,111 @@
...
+- job_name: "kube-scheduler"
+  scheme: http
+  honor_timestamps: true
+  metrics_path: /metrics
+  follow_redirects: true
+  enable_http2: true
+  static_configs:
+  - targets:
+    - "{{ .ProjectName }}-kube-scheduler:10251"
+{{ end }}
```
yibozhuang, line 111: new line.

pkg/kwokctl/utils/cmd.go
```diff
@@ -0,0 +1,166 @@
...
+		return fmt.Errorf("find process %d: %w", pid, err)
+	}
+	err = process.Kill()
+	if err != nil {
+		if errors.Is(err, os.ErrProcessDone) {
+			return nil
+		}
+		return fmt.Errorf("kill process: %w", err)
+	}
+	process.Wait()
```
yibozhuang, line 164: need to handle error here.

pkg/kwokctl/utils/cmd.go
```diff
@@ -0,0 +1,166 @@
...
+	"strings"
+)
+
+func ForkExec(ctx context.Context, dir string, name string, arg ...string) error {
+	pidPath := PathJoin(dir, "pids", filepath.Base(name)+".pid")
+	pidData, err := os.ReadFile(pidPath)
+	if err == nil {
+		_, err = strconv.Atoi(string(pidData))
+		if err == nil {
+			return nil // already running
```
yibozhuang, line 37: should you also call `os.FindProcess()` to ensure that it is already running?

otherwise there might be a case where if the kill failed and we do not remove the pid and that would cause inconsistencies.

pkg/kwokctl/utils/cmd.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return fmt.Errorf("read pid file %s: %w", pidPath, err)
+	}
+	pid, err := strconv.Atoi(string(raw))
+	if err != nil {
+		return fmt.Errorf("parse pid file %s: %w", pidPath, err)
+	}
+	err = killProcess(ctx, pid)
+	if err != nil {
+		return err
```
yibozhuang, line 113: so if the kill failed here, we would not remove the pid path, do we want to have this file remaining?

pkg/kwokctl/utils/download.go
```diff
@@ -0,0 +1,192 @@
...
+			}
+			return "", false
+		})
+		if err != nil {
+			return err
+		}
+		if clean {
+			os.Remove(cacheTar)
+		}
+		os.Chmod(cache, mode)
```
yibozhuang, line 59: need to handle error here.

pkg/kwokctl/utils/cmd.go
wzshiming: Yes
In `killProcess`, kill process success and process not exist should return nil, 
if it returns an error I think we should keep the pid path for later.

pkg/kwokctl/pki/admin.crt
wzshiming: Is the check format legal?

yibozhuang: I mean like, do these files need to be part of the PR? Since they are generated.

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
```diff
@@ -0,0 +1,202 @@
...
+			PrometheusBinaryTar:         flags.PrometheusBinaryTar,
+			CacheDir:                    vars.CacheDir,
+			SecretPort:                  flags.SecurePort,
+			QuietPull:                   flags.QuietPull,
+			PrometheusPort:              flags.PrometheusPort,
+			FeatureGates:                flags.KubeFeatureGates,
+			RuntimeConfig:               flags.KubeRuntimeConfig,
+		})
+		if err != nil {
+			return fmt.Errorf("failed install cluster %q: %w", name, err)
```
Huang-Wei, line 180: 💭 to brainstorm: should we cleanup the leftover once an error happens? How do other tools (like kind) deal with this situation?

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
```diff
@@ -0,0 +1,202 @@
...
+'${KWOK_KUBE_BINARY_PREFIX}/kube-controller-manager'
+`)
+	cmd.Flags().StringVar(&flags.KubeSchedulerBinary, "kube-scheduler-binary", vars.KubeSchedulerBinary, `binary of kube-scheduler, only for binary runtime
+'${KWOK_KUBE_BINARY_PREFIX}/kube-scheduler'
+`)
+	cmd.Flags().StringVar(&flags.KwokControllerBinary, "kwok-controller-binary", vars.KwokControllerBinary, `binary of kwok-controller, only for binary runtime
+`)
+	cmd.Flags().StringVar(&flags.EtcdBinaryTar, "etcd-binary-tar", vars.EtcdBinaryTar, `tar of etcd, only for binary runtime
+`)
+	cmd.Flags().StringVar(&flags.PrometheusBinaryTar, "prometheus-binary-tar", vars.PrometheusBinaryTar, `tar of Prometheus, only for binary runtime
```
Huang-Wei, line 109: Is `tar` intentional? I think we should accept binary path as the input, either to replace `tar` or coexist.

And an 💭 is to lookup the default path of etcd/prometheus and use it by default, if not specified.

Huang-Wei: for Mac M1, there is no prebuild etcd image/binary lol...

https://github.com/etcd-io/etcd/releases/tag/v3.5.4

- [etcd-v3.5.4-darwin-amd64.zip](https://github.com/etcd-io/etcd/releases/download/v3.5.4/etcd-v3.5.4-darwin-amd64.zip)
- [etcd-v3.5.4-linux-amd64.tar.gz](https://github.com/etcd-io/etcd/releases/download/v3.5.4/etcd-v3.5.4-linux-amd64.tar.gz)
- [etcd-v3.5.4-linux-arm64.tar.gz](https://github.com/etcd-io/etcd/releases/download/v3.5.4/etcd-v3.5.4-linux-arm64.tar.gz)
- [etcd-v3.5.4-linux-ppc64le.tar.gz](https://github.com/etcd-io/etcd/releases/download/v3.5.4/etcd-v3.5.4-linux-ppc64le.tar.gz)
- [etcd-v3.5.4-linux-s390x.tar.gz](https://github.com/etcd-io/etcd/releases/download/v3.5.4/etcd-v3.5.4-linux-s390x.tar.gz)
- [etcd-v3.5.4-windows-amd64.zip](https://github.com/etcd-io/etcd/releases/download/v3.5.4/etcd-v3.5.4-windows-amd64.zip)

pkg/kwokctl/utils/cmd_unix.go
```diff
@@ -0,0 +1,52 @@
+//go:build !windows
```
Huang-Wei, line 1: should we name this file to replace `cmd.go`? Or, two files named "cmd_other.go" and "cmd_windows.go".

pkg/kwokctl/utils/filepath_unix.go
```diff
@@ -0,0 +1,27 @@
+//go:build !windows
```
Huang-Wei, line 1: given it's non-windows, I'd suggest to rename it as `filepath_other.go`.

Huang-Wei (commented):
Thanks for this significant PR.

First found of review, but overall it's superb! Will test it locally after the "etcd binary path" parameter gets introduced (or "binary lookup logic" can be introduced).

pkg/kwokctl/runtime/registry.go
```diff
@@ -0,0 +1,67 @@
...
+	rt, ok = registry[name]
+	if ok {
+		return rt, true
+	}
+
+	return nil, false
+}
+
+func Load(name, workdir string, logger logger.Logger) (Runtime, error) {
+	dc := NewCluster(name, workdir, logger)
```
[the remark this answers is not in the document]
wzshiming: I think it's `docker compose`, which is the acronym I used for my first runtime implementation, and it's been copied ever since. I'll update it.

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
```diff
@@ -0,0 +1,202 @@
...
+			return err
+		}
+
+		err = dc.Down(ctx)
+		if err != nil {
+			logger.Printf("Failed to down cluster %q: %v", name, err)
+		}
+	} else {
+		logger.Printf("Creating cluster %q", name)
+		err = dc.Install(ctx, runtime.Config{
```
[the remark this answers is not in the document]
wzshiming: Failed to download binary or pull an image, or ran out of disk space.

pkg/kwokctl/runtime/kind/cluster.go
```diff
@@ -0,0 +1,285 @@
...
+	"strings"
+	"time"
+
+	"sigs.k8s.io/kwok/pkg/kwokctl/runtime"
+	"sigs.k8s.io/kwok/pkg/kwokctl/utils"
+	"sigs.k8s.io/kwok/pkg/logger"
+)
+
+type Cluster struct {
+	*runtime.Cluster
```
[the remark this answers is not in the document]
wzshiming: Yes
`*runtime.Cluster` is a common implementation of the interface by different Runtimes, such as saving and loading the current cluster configuration, checking if the cluster is ready.
I think it's easy to misunderstand, I'll move it to `runtime/common`.

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+
+		logger.Printf("Cluster %q is not ready yet, will be restarted", name)
+		err = dc.Install(ctx, *conf)
+		if err != nil {
+			logger.Printf("Failed to continue install cluster %q: %v", name, err)
+			return err
+		}
+
+		err = dc.Down(ctx)
```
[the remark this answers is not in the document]
wzshiming: The only reason to get here is that the cluster already exists, but is not ready for various reasons
`Install` is for an interruption halfway through the installation and then needs to continue the installation.
`Down` is for an interruption halfway through startup and then shutting down the already started one.

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
wzshiming: Is tar intentional? I think we should accept binary path as the input, either to replace tar or coexist.
Etcd and Prometheus releases do not provide a binary

        And an 💭 is to lookup the default path of etcd/prometheus and use it by default, if not specified.

That makes sense, I'll see how to do it right.

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
wzshiming: Kind no cleanup after a failed cluster creation, to save the scene for troubleshooting.

pkg/kwokctl/runtime/kind/cluster.go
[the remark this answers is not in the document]
Huang-Wei: ok, I understood it now. `Cluster` has different implementations on different "runtime".

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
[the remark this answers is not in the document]
Huang-Wei: Upon these errors, shouldn't we abort and quit? Otherwise, (re-)install would fail again?

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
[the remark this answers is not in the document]
Huang-Wei: > The only reason to get here

IIUC, when rt.Config(), rt.Ready(ctx) and rt.Install(ctx, *conf) all runs well and return nil err, the logic Down() is hit. so literally, it sounds to me it's a happy path, and why we call `Down()` in a happy path.

May there is something missing in the semantics of `err`. If so, I'd suggest to refactor the `err` as semantical `status`, so the readers can easily understand and distinguish the cases like "cluster already configured", etc.

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
Huang-Wei: We can do it in a follow-up.

Pretty nice, I can install it on my M1 🎉 :

```
⇒  kwokctl create cluster --runtime binary --kube-scheduler-binary /usr/local/bin/kube-scheduler@1.24.3 --kube-controller-manager-binary /usr/local/bin/kube-controller-manager@1.24.3  --kube-apiserver-binary /usr/local/bin/kube-apiserver@1.24.3 --kwok-controller-binary /usr/local/bin/kwok --etcd-binary /opt/homebrew/bin/etcd
Creating cluster "kwok-kwok"
Starting cluster "kwok-kwok"
Cluster "kwok-kwok" is ready
You can now use your cluster with:

    kubectl config use-context kwok-kwok

Thanks for using kwok!
```

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
Huang-Wei: My $0.02 is that it makes more sense to default to "cleanup upon faulty install". Take my UX for example, yesterday I installed it successfully, and it has a "cluster leftover". And today when I retry (with another sets of parameters), it seems to pickup the old config and failed anyway. It takes me some time to figure out the "leftover config was read" and I have to manually delete it.

cc @yibozhuang for thoughts.

pkg/kwokctl/cmd/get/get.go
```diff
@@ -0,0 +1,46 @@
...
+		Args:  cobra.NoArgs,
+		Use:   "get",
+		Short: "Gets one of [artifacts, clusters, kubeconfig]",
+		Long:  "Gets one of [artifacts, clusters, kubeconfig]",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return fmt.Errorf("subcommand is required")
+		},
+	}
+	// add subcommands
+	cmd.AddCommand(clusters.NewCommand(logger))
```
Huang-Wei, line 42: seems `kwokctl get artifacts --name kwok` returns the default images (instead of binaries) that the cluster is using:

```
⇒  kwokctl get artifacts --name kwok
https://dl.k8s.io/release/v1.24.2/bin/darwin/arm64/kubectl
registry.k8s.io/etcd:3.5.4-0
registry.k8s.io/kube-apiserver:v1.24.2
registry.k8s.io/kube-controller-manager:v1.24.2
registry.k8s.io/kube-scheduler:v1.24.2
registry.k8s.io/kwok/kwok:v0.1.0
docker.io/prom/prometheus:v2.35.0
```

The command I used to spin up the cluster was:

```
kwokctl create cluster --runtime binary --kube-scheduler-binary /usr/local/bin/kube-scheduler@1.24.3 --kube-controller-manager-binary /usr/local/bin/kube-controller-manager@1.24.3  --kube-apiserver-binary /usr/local/bin/kube-apiserver@1.24.3 --kwok-controller-binary /usr/local/bin/kwok --etcd-binary /opt/homebrew/bin/etcd
```

pkg/kwokctl/cmd/logs/logs.go
```diff
@@ -0,0 +1,72 @@
...
+		Args:  cobra.ExactArgs(1),
+		Use:   "logs",
+		Short: "Logs one of [etcd, kube-apiserver, kube-controller-manager, kube-scheduler, kwok-controller, prometheus]",
+		Long:  "Logs one of [etcd, kube-apiserver, kube-controller-manager, kube-scheduler, kwok-controller, prometheus]",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			flags.Name = vars.DefaultCluster
+			return runE(cmd.Context(), logger, flags, args)
+		},
+	}
+	cmd.Flags().BoolVarP(&flags.Follow, "follow", "f", false, "follow the log")
```
Huang-Wei, line 50:
```suggestion
	cmd.Flags().BoolVarP(&flags.Follow, "follow", "f", false, "Specify if the logs should be streamed.")
```

pkg/kwokctl/cmd/delete/delete.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/kwok/pkg/kwokctl/cmd/delete/cluster"
+	"sigs.k8s.io/kwok/pkg/logger"
+)
+
+// NewCommand returns a new cobra.Command for cluster creation
+func NewCommand(logger logger.Logger) *cobra.Command {
```
Huang-Wei, line 29: FYI: logs when deleting a running cluster using binary as the runtime:

```
⇒  kwokctl delete cluster
Stopping cluster "kwok-kwok"
failed to kill kwok: wait: no child processes
failed to kill kube-scheduler: wait: no child processes
failed to kill kube-controller-manager: wait: no child processes
failed to kill kube-apiserver: wait: no child processes
failed to kill etcd: wait: no child processes
Deleting cluster "kwok-kwok"
Cluster "kwok-kwok" deleted
```

Are the "failed to kill" strings as expected?

wzshiming:
Waiting merged of kubernetes-sigs/kwok#11 and rebase

pkg/kwokctl/pki/admin.crt
yibozhuang: @wzshiming just wanted to clarify if the cert files need to be part of the PR if we can generate them?

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
yibozhuang: Personally I would prefer to have the clean up done on a failed install but keep the log around for users to check.
Additionally, having a switch/flag to allow the user to run the install again without clean up for further debug.

We should be as idempotent as possible meaning that a previously failed case should ideally not affect future retries.

pkg/kwokctl/pki/admin.crt
wzshiming: Got it, I feel that adding an extra dependency on OpenSSL is not good, we can add a Go implementation later.

pkg/kwokctl/runtime/binary/promehteus.go
```diff
@@ -0,0 +1,51 @@
+/*
```
yibozhuang, line 1: typo for this filename: `prometheus.go`

pkg/kwokctl/runtime/list.go
```diff
@@ -0,0 +1,39 @@
...
+// ListClusters returns the list of clusters in the directory
+func ListClusters(workdir string) ([]string, error) {
+	entries, err := os.ReadDir(workdir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, err
+	}
+	ret := []string{}
```
yibozhuang, line 32:
```suggestion
	var ret []string
```

pkg/kwokctl/utils/cmd.go
```diff
@@ -0,0 +1,175 @@
...
+type IOStreams struct {
+	// In think, os.Stdin
+	In io.Reader
+	// Out think, os.Stdout
+	Out io.Writer
+	// ErrOut think, os.Stderr
+	ErrOut io.Writer
+}
+
+func killProcess(ctx context.Context, pid int) error {
```
yibozhuang, line 155: remove `ctx` since its not being used

pkg/kwokctl/utils/cmd_other.go
```diff
@@ -0,0 +1,52 @@
...
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+func startProcess(ctx context.Context, name string, arg ...string) *exec.Cmd {
+	cmd := command(ctx, name, arg...)
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		// Setsid is used to detach the process from the parent (normally a shell)
+		Setsid: true,
```
yibozhuang, line 32: is there a reason we set the sysproc attribute in `startProcess` here where as in `cmd_windows.go` the sysproc attribute for hide window is done in `command`? could we make it consistent?

pkg/kwokctl/utils/filepath_windows.go
```diff
@@ -0,0 +1,28 @@
...
+
+package utils
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+func PathJoin(elem ...string) string {
+	return strings.Replace(filepath.Join(elem...), `\`, `/`, -1)
```
yibozhuang, line 27: I'm confused on this: we want to modify the path because these are paths used within Docker/KinD running on windows correct?

What about user specifying `binary` when running on windows?

Makefile
```diff
@@ -40,5 +40,9 @@ integration-test:
 	@echo "Not implemented yet"
 
 .PHONY: e2e-test
-e2e-test:
-	@echo "Not implemented yet"
+e2e-test: \
```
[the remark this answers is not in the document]
Huang-Wei: nit: I think `.DEFAULT_GOAL := all` is the default setting already.

pkg/kwokctl/cmd/get/get.go
wzshiming: Fixed

pkg/kwokctl/utils/cmd_other.go
wzshiming: This is the attribute I got after independent testing on both sides, the definition of the sysproc attribute of the two systems is different

pkg/kwokctl/utils/filepath_windows.go
wzshiming: > I'm confused on this: we want to modify the path because these are paths used within Docker/KinD running on windows correct?

Yes

> What about user specifying binary when running on windows?

Both `\` and `/` work on Windows.

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
wzshiming: I added logic that will clean up upon faulty install.

pkg/kwokctl/cmd/kwokctl/create/cluster/cluster.go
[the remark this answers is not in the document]
wzshiming: Updated

wzshiming:
Updated for all comments
entrypoint of e2e-test will be modified after this merge, in #11

/unhold

yibozhuang (approved):

Huang-Wei:
/lgtm

We can fix nits in follow-ups. :shipit:

PR #6, opened 2022-08-03
Add Yibo to OWNERS

Huang-Wei:
Yibo helped a lot during incubating this project.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -3,3 +3,4 @@
 approvers:
   - Huang-Wei
   - wzshiming
+  - yibozhuang


wzshiming:
One way to make the user trusted is to add them as [members](https://git.k8s.io/community/community-membership.md#member) of the kubernetes-sigs org.

/lgtm
/hold

Maybe it should be left to trigger after Yibo becomes a member, in case something goes wrong with the automation process `/unhold`

wzshiming:
/unhold

It seems that my fears were superfluous which has been taken into account `do-not-merge/invalid-owners-file`

Huang-Wei:
@wzshiming you're right, @yibozhuang is working on a k8s member request.

wzshiming (approved):

PR #7, opened 2022-08-04
Makefile to organize CI jobs

Huang-Wei:
Once https://github.com/kubernetes/test-infra/pull/27039 gets merged, the basic CI setting should be effective.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,44 @@
+# Copyright 2022 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,44 @@
...
+	@echo "Verify gofmt"
+	@out=`gofmt -l -d $$(find . -name '*.go')`; \
+	if [ -n "$$out" ]; then \
+		echo "$$out"; \
+		exit 1; \
+	fi
+
+.PHONY: build
+build:
+	$(GO_CMD) build -o bin/kwok ./cmd/controller/*.go
```
wzshiming, line 36: `mkdir bin`
Missed creating bin directory

wzshiming: https://github.com/kubernetes-sigs/kwok/blob/0bf07fecd208ebaafadef7857dcb4d6bab606472/images/kwok/Dockerfile#L23
The binary built in `Dockerfile` is `controller`, which is `kwok` and I think it will cause unnecessary misunderstandings, I'll change it to `kwok` uniformly?

Huang-Wei: > Missed creating bin directory

It's created automatically :)

> I'll change it to kwok uniformly

Yup, let's use `kwok` as the binary name, and `kwok/kwok` as the image name in the future.

yibozhuang (approved):

Issue #2, opened 2022-08-02
[kwok] Evaluate the other promising contextual logger


wzshiming:
let's create an issue to evaluate the other promising contextual logger.

Kwok should probably output a structured log, Maybe use that implements the https://github.com/go-logr/logr or https://golang.org/x/exp/slog
Kwokctl should probably output a friendly log

_Originally posted by @Huang-Wei in https://github.com/kubernetes-sigs/kwok/pull/1#discussion_r935138322_

d3vus:
Hi @wzshiming, can i know the status of this issue - is it open or merged?

wzshiming:
It's still open

PR #12, opened 2022-08-10
Fix logger fmt

sivchari:
I fixed logger format.

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -17,7 +17,8 @@ limitations under the License.
 package logger
 
 type Logger interface {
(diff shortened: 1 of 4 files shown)


pkg/logger/logger.go
```diff
@@ -17,12 +17,14 @@ limitations under the License.
 package logger
 
 type Logger interface {
-	Printf(format string, args ...interface{})
+	Printf(format string, args ...any)
+	Println(v ...any)
```
wzshiming, line 21: Logger will be refactored in #2

wzshiming:
The logger library has been updated now, so let me close this. Thank you.
/close

PR #20, opened 2022-08-22
Release on github with gh-action

wzshiming:
I'm going to put a tag `v0.0.1-alpha.0` on this branch and see if the GH action can be triggered

/hold
/cc @Huang-Wei @yibozhuang

### DIFF
--- a/.github/workflows/releases.yaml
+++ b/.github/workflows/releases.yaml
@@ -1,0 +1,51 @@
+name: Releases
+on:
+  push:
+    tags:
+      - v*
+
+env:
+  PUSH: 'true'
+  DIRECT: 'true'
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  releases:
+    runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request'
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.18
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Log into registry ${{ env.REGISTRY }}
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build binary
+        env:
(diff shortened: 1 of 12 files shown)


.github/workflows/releases.yaml
```diff
@@ -0,0 +1,46 @@
+name: Releases
+on:
+  push:
+    tags:
+      - v*
+
+env:
+  PUSH: 'true'
+  DIRECT: 'true'
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
Huang-Wei, line 10: seems it's not used?

.github/workflows/releases.yaml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.18
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Log into registry
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
```
Huang-Wei, line 26: qq: does GH provide a native directive for login? Not sure this plain `run` directive may leak the token? like display in the log publicly accessible?

Huang-Wei (commented):
some nits.

.github/workflows/releases.yaml
wzshiming: `gh release upload` will use `GH_TOKEN` to put `binary` into the release page.

.github/workflows/releases.yaml
wzshiming: Updated

.github/workflows/releases.yaml
```diff
@@ -0,0 +1,51 @@
...
+  releases:
+    runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request'
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.18
+      - name: Set up QEMU
```
yibozhuang, line 22: QEMU is for cross-compile correct?

yibozhuang (approved):

.github/workflows/releases.yaml
wzshiming: Is for building multi-architecture images.

wzshiming: https://github.com/docker/setup-qemu-action#about

PR #21, opened 2022-08-23
Use scripts to maintain the supported_releases.txt file

wzshiming:
/kind feature
/hold

### DIFF
--- a/hack/supported_releases.sh
+++ b/hack/supported_releases.sh
@@ -1,0 +1,119 @@
+#!/usr/bin/env bash
+# Copyright 2022 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


yibozhuang (approved):
LGTM

1 question: when should we run this script to update the supported_releases file? Manually? This is tracking the kind base image versions so I guess we'd need to update it whenever there is a new kind image.

Manually

wzshiming:
@yibozhuang 
Update
Added timed task to check daily, and send PR if out of sync

#24 Run in my local, that look good.

hack/supported_releases.sh
```diff
@@ -0,0 +1,116 @@
...
+    return 1
+  fi
+  if [[ "${status}" != " M supported_releases.txt" ]]; then
+    echo "Modified unintended documents"
+    echo "${status}"
+    return 1
+  fi
+
+  # Use the latest release as the key to prevent duplicate PRs from being created
+  latest="$(cat supported_releases.txt | head -n 1)"
```
yibozhuang, line 71: this seems hacky, what if a previous release was updated?

i.e. latest release 1.25 did not change but 1.24 was updated?

maybe use the timestamp here as the key to prevent duplicate?

yibozhuang (commented):
one nit, rest looks good

hack/supported_releases.sh
wzshiming: If we use the timestamp as the key, then the PR created on the first day without processing will be created again the next day, 
or just use a fixed branch, so that no matter what, it will not be created again.

yibozhuang (approved):
/lgtm

PR #27, opened 2022-08-26
Modify the default working directory

wzshiming:
Modify the default working directory in Home, and add the KWOK_WORKDIR to configure it, avoid accidental deletion.

/kind feature


yibozhuang (approved):
/lgtm

PR #30, opened 2022-08-29
Add e2e for darwin and linux on gh-action

wzshiming:


.github/workflows/test.yaml
```diff
@@ -0,0 +1,108 @@
...
+        shell: bash
+        run: |
+          ./hack/e2e-test.sh kwokctl/kwokctl_${{ matrix.kwokctl-runtime }}_snapshot
+
+      - name: Test Benchmark
+        shell: bash
+        if: matrix.kwokctl-runtime != 'kind'
+        continue-on-error: true
+        run: |
+          ./hack/e2e-test.sh kwokctl/kwokctl_${{ matrix.kwokctl-runtime }}_benchmark
```
Huang-Wei, line 108: Is that we're going to use GH action test kwokctl, and use k8s test-infra to test kwok?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,108 @@
...
+        shell: bash
+        run: |
+          ./hack/requirements.sh kind
+          kind version
+
+      - name: Docker Info
+        if: ${{ matrix.kwokctl-runtime != 'binary' }}
+        shell: bash
+        run: |
+          docker info
```
yibozhuang, line 85: what's the purpose of this? just for logging the docker version?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+      - name: Test Snapshot
+        shell: bash
+        run: |
+          ./hack/e2e-test.sh kwokctl/kwokctl_${{ matrix.kwokctl-runtime }}_snapshot
+
+      - name: Test Benchmark
+        shell: bash
+        if: matrix.kwokctl-runtime != 'kind'
+        continue-on-error: true
```
yibozhuang, line 106: why continue on error here and not the other tests?

.github/workflows/test.yaml
wzshiming: Ooops I seem to have missed Kwok's e2e

.github/workflows/test.yaml
wzshiming: This often times out in k8s test-infra, depending on machine performance

.github/workflows/test.yaml
wzshiming: Yes

liangyuanpeng (approved):

PR #37, opened 2022-09-02
Add KWOK_MODE to configure several default cluster templates

wzshiming:
/kind feature
/kind bug

Fixes #36

### DIFF
--- a/pkg/kwokctl/vars/vars.go
+++ b/pkg/kwokctl/vars/vars.go
@@ -171,11 +171,28 @@ var (
 		return ""
 	}()
 
+	// Mode is several default parameter templates for clusters
+	Mode = getEnv("KWOK_MODE", "")
+
+	// ModeStableFeatureGateAndAPI efforts to reduce cluster configuration requirements
+	// Disables all alpha feature by default, as well as Bata feature that are not eventually GA
+	ModeStableFeatureGateAndAPI = "StableFeatureGateAndAPI"
+
 	// KubeFeatureGates is a set of key=value pairs that describe feature gates for alpha/experimental features of Kubernetes.
-	KubeFeatureGates = getEnv("KWOK_KUBE_FEATURE_GATES", k8s.GetFeatureGates(parseRelease(KubeVersion)))
+	KubeFeatureGates = getEnv("KWOK_KUBE_FEATURE_GATES", func() string {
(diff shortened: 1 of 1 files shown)


pkg/kwokctl/vars/vars.go
```diff
@@ -171,11 +171,28 @@ var (
 		return ""
 	}()
 
+	// Mode is several default parameter templates for clusters
+	Mode = getEnv("KWOK_MODE", "")
+
+	// ModeTiny efforts to reduce cluster configuration requirements
+	// Disables all alpha feature by default, as well as Bata feature that are not eventually GA
+	ModeTiny = "tiny"
```
Huang-Wei, line 179: "tiny" doesn't read well, can we change to another name? like `StableFeatureFateAndAPI`?

pkg/kwokctl/vars/vars.go
```diff
@@ -171,11 +171,28 @@ var (
 		return ""
 	}()
 
+	// Mode is several default parameter templates for clusters
+	Mode = getEnv("KWOK_MODE", "")
```
Huang-Wei, line 175: So the default mode would be "", is it intentional?

wzshiming: Yes, No special configuration is done by default.

pkg/kwokctl/vars/vars.go
wzshiming: Updated

pkg/kwokctl/vars/vars.go
```diff
@@ -171,11 +171,28 @@ var (
...
+
+	// ModeStableFeatureGateAndAPI efforts to reduce cluster configuration requirements
```
yibozhuang, line 177:
```suggestion
	// ModeStableFeatureGateAndAPI is intended to reduce cluster configuration requirements
```

pkg/kwokctl/vars/vars.go
```diff
@@ -171,11 +171,28 @@ var (
...
+	// Disables all alpha feature by default, as well as Bata feature that are not eventually GA
```
yibozhuang, line 178:
```suggestion
	// Disables all Alpha feature by default, as well as Beta feature that are not eventually GA
```

yibozhuang (approved):
/lgtm

PR #42, opened 2022-09-06
🐛reuse kubectl found in path

Huang-Wei:
In my M1 env, I'd like to reuse the `kubectl`; otherwise you cannot download a darwin/arm64 kubectl binary:

w/o this PR, the following command won't work in M1:

```
KWOK_KUBE_VERSION=v1.20.15 kwokctl create cluster --runtime binary --kube-scheduler-binary /usr/local/bin/kube-scheduler@1.20.15 --kube-controller-manager-binary /usr/local/bin/kube-controller-manager@1.20.15  --kube-apiserver-binary /usr/local/bin/kube-apiserver@1.20.15 --kwok-controller-binary /usr/local/bin/kwok --etcd-binary /opt/homebrew/bin/etcd --name 120-binary
```

### DIFF
--- a/pkg/kwokctl/runtime/cluster.go
+++ b/pkg/kwokctl/runtime/cluster.go
@@ -193,12 +193,16 @@ func (c *Cluster) KubectlInCluster(ctx context.Context, stm utils.IOStreams, arg
 		return err
 	}
 
-	bin := utils.PathJoin(conf.Workdir, "bin")
-	kubectlPath := utils.PathJoin(bin, "kubectl"+vars.BinSuffix)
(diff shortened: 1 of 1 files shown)


wzshiming:
Ooops seems to have triggered another bug, I've fixed it, #44 merge first and then rebase this RP

wzshiming:
Do we need to provide a way to disable kubectl in the path? Otherwise it will prompt
>WARNING: version difference between client (1.24) and server (1.22) exceeds the supported minor version skew of +/-1

Huang-Wei:
> Do we need to provide a way to disable kubectl in the path? Otherwise it will prompt

It doesn't bother me as kubectl version skew is more robust than the skew among other core components.

PR #44, opened 2022-09-07
Fix symlink

wzshiming:


Huang-Wei:
@wzshiming We may need to disable part of the tests in `e2e-test-main`, otherwise it's quite flaky.

yibozhuang (approved):
/lgtm

PR #53, opened 2022-10-28
add comment note for node-label-selector filter logic

yanggangtony:
The parameter manageNodesWithLabelSelector is used in client-go opt filter logic , while the parameter manageNodesWithAnnotationSelector is in NodeController's filter nodeSelectorFunc logic, that confused me..

### DIFF
--- a/pkg/controllers/controller.go
+++ b/pkg/controllers/controller.go
@@ -95,6 +95,8 @@ func NewController(conf Config) (*Controller, error) {
 		nodeSelectorFunc = func(node *corev1.Node) bool {
 			return selector.Matches(labels.Set(node.Annotations))
 		}
+		// client-go supports label filtering, so ignore this.
+		// } else if conf.ManageNodesWithLabelSelector != "" {
 	}
 
 	var lockPodsOnNodeFunc func(ctx context.Context, nodeName string) error


wzshiming:
good catch
/ok-to-test

pkg/controllers/controller.go
```diff
@@ -95,6 +95,14 @@ func NewController(conf Config) (*Controller, error) {
...
 			return selector.Matches(labels.Set(node.Annotations))
 		}
+	} else if conf.ManageNodesWithLabelSelector != "" {
+		selector, err := labels.Parse(conf.ManageNodesWithLabelSelector)
+		if err != nil {
+			return nil, err
+		}
+		nodeSelectorFunc = func(node *corev1.Node) bool {
+			return selector.Matches(labels.Set(node.Labels))
+		}
```
wzshiming, line 105:
```suggestion
    // client-go supports label filtering, so ignore this.
	// } else if conf.ManageNodesWithLabelSelector != "" {
```

https://github.com/kubernetes-sigs/kwok/blob/a830f47bfb036704093ed06b5caf79c3c1723d5b/pkg/controllers/node_controller.go#L135

`client-go` supports a label selector, so this is not necessary, but we can add a comment to clarify it

yanggangtony:
@wzshiming thanks , just now i saw the code ..
it is used in opt, so it is not  so this is not necessary

PR #70, opened 2022-11-17
subcommand silence usage and parent command print usage instead of print err

carlory:
/kind feature

### DIFF
--- a/pkg/kwok/cmd/root.go
+++ b/pkg/kwok/cmd/root.go
@@ -65,10 +65,11 @@ func NewCommand(logger logger.Logger) *cobra.Command {
 	)
 
 	cmd := &cobra.Command{
-		Args:  cobra.NoArgs,
-		Use:   "kwok [command]",
-		Short: "kwok is a tool for simulate thousands of fake kubelets",
-		Long:  "kwok is a tool for simulate thousands of fake kubelets",
+		Args:         cobra.NoArgs,
+		Use:          "kwok [command]",
+		Short:        "kwok is a tool for simulate thousands of fake kubelets",
+		Long:         "kwok is a tool for simulate thousands of fake kubelets",
+		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if kubeconfig != "" {
 				f, err := os.Stat(kubeconfig)
--- a/pkg/kwokctl/cmd/create/cluster/cluster.go
+++ b/pkg/kwokctl/cmd/create/cluster/cluster.go
@@ -63,10 +63,11 @@ type flagpole struct {
 func NewCommand(logger logger.Logger) *cobra.Command {
 	flags := &flagpole{}
 	cmd := &cobra.Command{
-		Args:  cobra.NoArgs,
-		Use:   "cluster",
-		Short: "Creates a cluster",
-		Long:  "Creates a cluster",
+		Args:         cobra.NoArgs,
+		Use:          "cluster",
+		Short:        "Creates a cluster",
+		Long:         "Creates a cluster",
+		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			flags.Name = vars.DefaultCluster
 			return runE(cmd.Context(), logger, flags)
--- a/pkg/kwokctl/cmd/create/create.go
+++ b/pkg/kwokctl/cmd/create/create.go
@@ -17,8 +17,6 @@ limitations under the License.
 package create
 
 import (
-	"fmt"
-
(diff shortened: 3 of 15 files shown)


wzshiming (changes requested):
There is actual output, but it is swiped by subsequent Usage messages, so maybe we can hide the Usage messages when the process executes with an error.

carlory:
It maybe necessary to print err in the main.go, because we does not make sure that error message is logged.

FYI: https://github.com/kubernetes-sigs/kwok/blob/main/pkg/kwok/cmd/root.go#L84


> so maybe we can hide the Usage messages when the process executes with an error.

It makes sense.

carlory:
cc @wzshiming

cmd/kwok/main.go
```diff
@@ -28,6 +29,7 @@ func main() {
 	command := cmd.NewCommand(logger)
 	err := command.Execute()
 	if err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
```
wzshiming, line 32:
```
W1118 11:00:42.067335   81145 client_config.go:617] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
W1118 11:00:42.067360   81145 client_config.go:622] error creating inClusterConfig, falling back to default config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
Error: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
error: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
exit status 1
```

wzshiming: Maybe this could be removed, or the built-in default error output could be disabled?

pkg/kwok/cmd/root.go
```diff
@@ -65,10 +65,11 @@ func NewCommand(logger logger.Logger) *cobra.Command {
...
 	cmd := &cobra.Command{
-		Args:  cobra.NoArgs,
-		Use:   "kwok [command]",
-		Short: "kwok is a tool for simulate thousands of fake kubelets",
-		Long:  "kwok is a tool for simulate thousands of fake kubelets",
+		Args:         cobra.NoArgs,
+		Use:          "kwok [command]",
+		Short:        "kwok is a tool for simulate thousands of fake kubelets",
+		Long:         "kwok is a tool for simulate thousands of fake kubelets",
+		SilenceUsage: true,
```
wzshiming, line 72: It seems to have missed some `cobra.Command`s that don't have this option, I counted 15 of them and we need to add this option.

cmd/kwok/main.go
carlory: we can remove it, because cobra will print error info when runE returns an error

pkg/kwok/cmd/root.go
carlory: for parent command, we can puts out the help for the command instead of print error. 


```

Error: subcommand is required
Usage:
  kwokctl [command] [flags]
  kwokctl [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  create      Creates one of [cluster]
  delete      Deletes one of [cluster]
  get         Gets one of [artifacts, clusters, kubeconfig]
  help        Help about any command
  kubectl     kubectl in cluster
  logs        Logs one of [audit, etcd, kube-apiserver, kube-controller-manager, kube-scheduler, kwok-controller, prometheus]
  snapshot    Snapshot [save, restore] one of cluster

Flags:
  -h, --help          help for kwokctl
      --name string   cluster name (default "kwok")

Use "kwokctl [command] --help" for more information about a command.

exit status 1
```

expected:

```
Kwokctl is a Kwok cluster management tool

Usage:
  kwokctl [command] [flags]
  kwokctl [command]

Available Commands:
  completion  Generate the autocompletion script for the specified shell
  create      Creates one of [cluster]
  delete      Deletes one of [cluster]
  get         Gets one of [artifacts, clusters, kubeconfig]
  help        Help about any command
  kubectl     kubectl in cluster
  logs        Logs one of [audit, etcd, kube-apiserver, kube-controller-manager, kube-scheduler, kwok-controller, prometheus]
  snapshot    Snapshot [save, restore] one of cluster

Flags:
  -h, --help          help for kwokctl
      --name string   cluster name (default "kwok")

Use "kwokctl [command] --help" for more information about a command.
```

carlory:
/retitle subcommand silence usage and parent command print usage instead of print err

carlory:
cc @wzshiming 

It‘s ready for review.

PR #74, opened 2022-11-18
Support Pod/Node lifecycle simulation

### LINKED ISSUE #67: [kwok] Pod lifecycle simulation
This is just a preliminary design

## Goals
- Support simulate the pod's real behavior as much as possible
- Flexible and configurable
- Present some templates:
  - Up and running quickly
  - Simulates realistic normal/abnormal behavior

## Design 1

<details>

``` yaml
kind: Scenario
metadata:
  name: scenario-0
spec:
  selector:
    matchLabels:
      app: demo
  stages:
  - kind: Duration
    delay: 5s
    jitter: 0.2

  - kind: Switch
    cases:
    - weight: 9
      stages: 
      - kind: PresetStatus
        template: PullImage
    - weight: 1
      stages: 
      - kind: PresetStatus
        template: FailedPullImage
      - kind: GoToBegin

  - kind: Duration
    delay: 1s

  - kind: PresetStatus
    template: Creating

  - kind: Duration
    delay: 1s
    jitter: 0.1

  - kind: PresetStatus
    template: NotReady

  - kind: Duration
    delay: 1s
    jitter: 0.1

  - kind: Switch
    cases:
    - weight: 20
      stages: 
      - kind: PresetStatus
        template: Ready
      - kind: End

    - weight: 5
      stages: 
      - kind: PresetStatus
        template: Failed
      - kind: End

    - weight: 5
      - kind: PresetStatus
        template: Restart
      - kind: GoToBegin

```

</details>

## Design 2

``` yaml
kind: Stage
metadata:
  name: pod-create
spec:
  resourceRef:
    apiGroup: v1
    kind: Pod
  selector:
    matchExpressions:
      - key: '.metadata.deletionTimestamp'
        operator: 'DoesNotExist'
      - key: '.status.podIP'
        operator: 'DoesNotExist'
  weight: 1
  delay:
    durationSeconds: 1
  next:
    event:
      type: Normal
      reason: Created
      message: Created container
    finalizers:
      add:
        - value: 'kwok.x-k8s.io/fake'
    statusTemplate: |
      {{ $now := Now }}
      conditions:
      {{ if .spec.initContainers }}
        - lastProbeTime: null
          lastTransitionTime: '{{ $now }}'
          message: 'containers with incomplete status: [{{ range .spec.initContainers }} {{ .name }} {{ end }}]'
          reason: ContainersNotInitialized
          status: "False"
          type: Initialized
      {{ else }}
        - lastProbeTime: null
          lastTransitionTime: '{{ $now }}'
          status: "True"
          type: Initialized
      {{ end }}
        - lastProbeTime: null
          lastTransitionTime: '{{ $now }}'
          message: 'containers with unready status: [{{ range .spec.containers }} {{ .name }} {{ end }}]'
          reason: ContainersNotReady
          status: "False"
          type: Ready
        - lastProbeTime: null
          lastTransitionTime: '{{ $now }}'
          message: 'containers with unready status: [{{ range .spec.containers }} {{ .name }} {{ end }}]'
          reason: ContainersNotReady
          status: "False"
          type: ContainersReady
      {{ range .spec.readinessGates }}
        - lastTransitionTime: {{ $now }}
          status: "True"
          type: {{ .conditionType }}
      {{ end }}
      {{ if .spec.initContainers }}
      initContainerStatuses:
        {{ range .spec.initContainers }}
        - image: {{ .image }}
          name: {{ .name }}
          ready: false
          restartCount: 0
          started: false
          state:
            waiting:
              reason: PodInitializing
        {{ end }}
      containerStatuses:
        {{ range .spec.containers }}
        - image: {{ .image }}
          name: {{ .name }}
          ready: false
          restartCount: 0
          started: false
          state:
            waiting:
              reason: PodInitializing
        {{ end }}
      {{ else }}
      containerStatuses:
        {{ range .spec.containers }}
        - image: {{ .image }}
          name: {{ .name }}
          ready: false
          restartCount: 0
          started: false
          state:
            waiting:
              reason: ContainerCreating
        {{ end }}
      {{ end }}
      hostIP: {{ with .status.hostIP }} {{ . }} {{ else }} {{ NodeIP }} {{ end }}
      podIP: {{ with .status.podIP }} {{ . }} {{ else }} {{ PodIP }} {{ end }}
      phase: Pending
---
kind: Stage
metadata:
  name: pod-init-container-running
spec:
  resourceRef:
    apiGroup: v1
    kind: Pod
  selector:
    matchExpressions:
      - key: '.metadata.deletionTimestamp'
        operator: 'DoesNotExist'
      - key: '.status.phase'
        operator: 'In'
        values:
          - 'Pending'
      - key: '.status.conditions.[] | select( .type == "Initialized" ) | .status'
        operator: 'NotIn'
        values:
          - 'True'
      - key: '.status.initContainerStatuses.[].state.waiting.reason'
        operator: 'Exists'
  weight: 1
  delay:
    durationSeconds: 5
  next:
    statusTemplate: |
      {{ $now := Now }}
      {{ $root := . }}
      initContainerStatuses:
        {{ range $index, $item := .spec.initContainers }}
        {{ $origin := index $root.status.initContainerStatuses $index }}
        - image: {{ $item.image }}
          name: {{ $item.name }}
          ready: true
          restartCount: 0
          started: true
          state:
            running:
              startedAt: '{{ $now }}'
        {{ end }}
---
kind: Stage
metadata:
  name: pod-init-container-completed
spec:
  resourceRef:
    apiGroup: v1
    kind: Pod
  selector:
    matchExpressions:
      - key: '.metadata.deletionTimestamp'
        operator: 'DoesNotExist'
      - key: '.status.phase'
        operator: 'In'
        values:
          - 'Pending'
      - key: '.status.conditions.[] | select( .type == "Initialized" ) | .status'
        operator: 'NotIn'
        values:
          - 'True'
      - key: '.status.initContainerStatuses.[].state.running.startedAt'
        operator: 'Exists'
  weight: 1
  delay:
    durationSeconds: 5
  next:
    statusTemplate: |
      {{ $now := Now }}
      {{ $root := . }}
      conditions:
        - lastProbeTime: null
          lastTransitionTime: '{{ $now }}'
          status: "True"
          type: Initialized
      initContainerStatuses:
        {{ range $index, $item := .spec.initContainers }}
        {{ $origin := index $root.status.initContainerStatuses $index }}
        - image: {{ $item.image }}
          name: {{ $item.name }}
          ready: true
          restartCount: 0
          started: false
          state:
            terminated:
              exitCode: 0
              finishedAt: '{{ $now }}'
              reason: Completed
              startedAt: '{{ $now }}'
        {{ end }}
      containerStatuses:
        {{ range .spec.containers }}
        - image: {{ .image }}
          name: {{ .name }}
          ready: false
          restartCount: 0
          started: false
          state:
            waiting:
              reason: ContainerCreating
        {{ end }}
---
kind: Stage
metadata:
  name: pod-ready
spec:
  resourceRef:
    apiGroup: v1
    kind: Pod
  selector:
    matchExpressions:
      - key: '.metadata.deletionTimestamp'
        operator: 'DoesNotExist'
      - key: '.status.phase'
        operator: 'In'
        values:
          - 'Pending'
      - key: '.status.conditions.[] | select( .type == "Initialized" ) | .status'
        operator: 'In'
        values:
          - 'True'
      - key: '.status.conditions.[] | select( .type == "ContainersReady" ) | .status'
        operator: 'NotIn'
        values:
          - 'True'
  weight: 1
  delay:
    durationSeconds: 10
  next:
    delete: false
    statusTemplate: |
      {{ $now := Now }}
      {{ $root := . }}
      conditions:
        - lastProbeTime: null
          lastTransitionTime: '{{ $now }}'
          message: ''
          reason: ''
          status: "True"
          type: Ready
        - lastProbeTime: null
          lastTransitionTime: '{{ $now }}'
          message: ''
          reason: ''
          status: "True"
          type: ContainersReady
      containerStatuses:
        {{ range $index, $item := .spec.containers }}
        {{ $origin := index $root.status.containerStatuses $index }}
        - image: {{ $item.image }}
          name: {{ $item.name }}
          ready: true
          restartCount: 0
          started: true
          state:
            running:
              startedAt: '{{ $now }}'
        {{ end }}
      phase: Running
---
kind: Stage
metadata:
  name: pod-completed-for-job
spec:
  resourceRef:
    apiGroup: v1
    kind: Pod
  selector:
    matchExpressions:
      - key: '.metadata.deletionTimestamp'
        operator: 'DoesNotExist'
      - key: '.status.phase'
        operator: 'In'
        values:
          - 'Running'
      - key: '.status.conditions.[] | select( .type == "Ready" ) | .status'
        operator: 'In'
        values:
          - 'True'
      - key: '.metadata.ownerReferences.[].kind'
        operator: 'In'
        values:
          - 'Job'
  weight: 1
  delay:
    durationSeconds: 10
  next:
    delete: false
    statusTemplate: |
      {{ $now := Now }}
      {{ $root := . }}
      containerStatuses:
        {{ range $index, $item := .spec.containers }}
        {{ $origin := index $root.status.containerStatuses $index }}
        - image: {{ $item.image }}
          name: {{ $item.name }}
          ready: true
          restartCount: 0
          started: false
          state:
            terminated:
              exitCode: 0
              finishedAt: '{{ $now }}'
              reason: Completed
              startedAt: '{{ $now }}'
        {{ end }}
      phase: Succeeded
---
kind: Stage
metadata:
  name: pod-remove-finalizer
spec:
  resourceRef:
    apiGroup: v1
    kind: Pod
  selector:
    matchExpressions:
      - key: '.metadata.deletionTimestamp'
        operator: 'Exists'
      - key: '.metadata.finalizers'
        operator: 'In'
        values:
          - 'kwok.x-k8s.io/fake'
  weight: 1
  delay:
    durationSeconds: 1
  next:
    finalizers:
      remove:
        - value: 'kwok.x-k8s.io/fake'
    event:
      type: Normal
      reason: Killing
      message: Stopping container
---
kind: Stage
metadata:
  name: pod-delete
spec:
  resourceRef:
    apiGroup: v1
    kind: Pod
  selector:
    matchExpressions:
      - key: '.metadata.deletionTimestamp'
        operator: 'Exists'
      - key: '.metadata.finalizers'
        operator: 'DoesNotExist'
  weight: 1
  delay:
    durationSeconds: 1
    jitterDurationSecondsFrom:
      expressionFrom: '.metadata.deletionTimestamp'
  next:
    delete: true
```

``` golang
package v1alpha1

import (
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
)

// Stage is an API that describes the staged change of a resource
type Stage struct {
	metav1.TypeMeta `json:",inline"`
	// Standard list metadata.
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
	metav1.ObjectMeta `json:"metadata,omitempty"`
	// Spec holds information about the request being evaluated.
	Spec StageSpec `json:"spec,omitempty"`
}

// StageSpec defines the specification for Stage.
type StageSpec struct {
	// ResourceRef specifies the Kind and version of the resource.
	ResourceRef StageResourceRef `json:"resourceRef"`
	// Selector specifies the stags will be applied to the selected resource.
	Selector *StageSelector `json:"selector,omitempty"`
	// Weight means the current stage, in case of multiple stages,
	// a random stage will be matched as the next stage based on the weight.
	Weight *int `json:"weight,omitempty"`
	// Delay means there is a delay in this stage.
	Delay *StageDelay `json:"delay,omitempty"`
	// Next indicates that this stage will be moved to.
	Next StageNext `json:"next"`
}

// StageResourceRef specifies the kind and version of the resource.
type StageResourceRef struct {
	// APIGroup of the referent.
	APIGroup *string `json:"apiGroup,omitempty"`
	// Kind of the referent.
	Kind string `json:"kind"`
}

// StageDelay describes the delay time before going to next.
type StageDelay struct {
	// DurationSeconds indicates the stage delay time (in seconds).
	// If JitterDurationSeconds is less than DurationSeconds, then JitterDurationSeconds is used.
	DurationSeconds int `json:"durationSeconds"`
	// DurationSecondsFrom is the expression used to get the value.
	// If it is a time.Time type, getting the value will be minus now() to get DurationSeconds
	// If it is a string type, the value get will be parsed by time.ParseDuration.
	DurationSecondsFrom *ExpressionFromSource `json:"durationSecondsFrom,omitempty"`

	// JitterDurationSeconds is the duration plus an additional amount chosen uniformly
	// at random from the interval between `durationSeconds` and `jitterDurationSeconds`.
	JitterDurationSeconds *int `json:"jitterDurationSeconds,omitempty"`
	// JitterDurationSecondsFrom is the expression used to get the value.
	// If it is a time.Time type, getting the value will be minus now() to get JitterDurationSeconds
	// If it is a string type, the value get will be parsed by time.ParseDuration.
	JitterDurationSecondsFrom *ExpressionFromSource `json:"jitterDurationSecondsFrom,omitempty"`
}

// StageNext describes a stage will be moved to.
type StageNext struct {
	// Event means that an event will be sent.
	Event *StageEvent `json:"event,omitempty"`
	// Finalizers means that finalizers will be modified.
	Finalizers *StageFinalizers `json:"finalizers,omitempty"`
	// Delete means that the resource will be deleted if true.
	Delete bool `json:"delete,omitempty"`
	// StatusTemplate indicates the template for modifying the status of the resource in the next.
	StatusTemplate string `json:"statusTemplate,omitempty"`
}

// StageFinalizers describes the modifications in the finalizers of a resource.
type StageFinalizers struct {
	// Add means that the Finalizers will be added to the resource.
	Add []FinalizerItem `json:"add,omitempty"`
	// Remove means that the Finalizers will be removed from the resource.
	Remove []FinalizerItem `json:"remove,omitempty"`
	// Empty means that the Finalizers for that resource will be emptied.
	Empty bool `json:"empty,omitempty"`
}

// FinalizerItem  describes the one of the finalizers.
type FinalizerItem struct {
	Value string `json:"value,omitempty"`
}

// StageEvent describes one event in the Kubernetes.
type StageEvent struct {
	// Type is the type of this event (Normal, Warning), It is machine-readable.
	Type string `json:"type,omitempty"`
	// Reason is why the action was taken. It is human-readable.
	Reason string `json:"reason,omitempty"`
	// Message is a human-readable description of the status of this operation.
	Message string `json:"message,omitempty"`
}

// StageSelector is a resource selector. the result of matchLabels and matchAnnotations and
// matchExpressions are ANDed. An empty resource selector matches all objects. A null
// resource selector matches no objects.
type StageSelector struct {
	// MatchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels
	// map is equivalent to an element of matchExpressions, whose key field is ".metadata.labels[key]", the
	// operator is "In", and the values array contains only "value". The requirements are ANDed.
	MatchLabels map[string]string `json:"matchLabels,omitempty"`
	// MatchAnnotations is a map of {key,value} pairs. A single {key,value} in the matchAnnotations
	// map is equivalent to an element of matchExpressions, whose key field is ".metadata.annotations[key]", the
	// operator is "In", and the values array contains only "value". The requirements are ANDed.
	MatchAnnotations map[string]string `json:"matchAnnotations,omitempty"`
	// MatchExpressions is a list of label selector requirements. The requirements are ANDed.
	MatchExpressions []SelectorRequirement `json:"matchExpressions,omitempty"`
}

// SelectorRequirement is a resource selector requirement is a selector that contains values, a key,
// and an operator that relates the key and values.
type SelectorRequirement struct {
	// The name of the scope that the selector applies to.
	Key string `json:"key"`
	// Represents a scope's relationship to a set of values.
	Operator SelectorOperator `json:"operator"`
	// An array of string values.
	// If the operator is In, NotIn, Intersection or NotIntersection, the values array must be non-empty.
	// If the operator is Exists or DoesNotExist, the values array must be empty.
	Values []string `json:"values,omitempty"`
}

// SelectorOperator is a label selector operator is the set of operators that can be used in a selector requirement.
type SelectorOperator string

var (
	SelectorOpIn           SelectorOperator = "In"
	SelectorOpNotIn        SelectorOperator = "NotIn"
	SelectorOpExists       SelectorOperator = "Exists"
	SelectorOpDoesNotExist SelectorOperator = "DoesNotExist"
)

// ExpressionFromSource represents a source for the value of a from.
type ExpressionFromSource struct {
	// ExpressionFrom is the expression used to get the value.
	ExpressionFrom string `json:"expressionFrom,omitempty"`
}
```

wzshiming:
Fixes #67

/kind feature


stages/pod-general.yaml
```diff
@@ -0,0 +1,357 @@
...
+          state:
+            waiting:
+              reason: ContainerCreating
+        {{ end }}
+      {{ end }}
+      hostIP: {{ with .status.hostIP }} {{ . }} {{ else }} {{ NodeIP }} {{ end }}
+      podIP: {{ with .status.podIP }} {{ . }} {{ else }} {{ PodIP }} {{ end }}
+      phase: Pending
+---
+kind: Stage
```
kerthcet, line 100: From my point of view, it would be a little too complex for end-users to config based on current implementations, providing examples mitigate this a little but not much I think, especially for someone who're SREs not developers.

Besides, I would like to see something more like this, but I have no idea whether this meets all your needs and descriptive enough. This can also not a CRD I think. 
``` yaml
Steps:
- name: create pod
   phase: Pending
   filePath: <path-to-pod.yaml>
   condition:
     type:
     status:
     lastProbeTime:
     lastTransitionTime:
     reason:
     message:
- name: running pod
   after: 10s # waiting time after prev step
   phase: Running
   condition:
- name: liveness probe error # something like this
   phase: Fail
   condition:
- name: delete pod
   phase: Terminating
   condition:
- name: pod is terminated
   phase: Terminated
# - phase: Succeed
# - phase: Unknown
```

This is declarative and if necessary, you can export pod field if you like.

wzshiming: https://github.com/kubernetes-sigs/kwok/issues/67
https://github.com/kubernetes-sigs/kwok/issues/67#issuecomment-1312732001

This is my first version of the design, which has the fatal problem of needing to have a state to maintain the stage where the Pod is, 
and if the kwok is restarted, it can't pick up where it left off.

stages/pod-fast.yaml
```diff
@@ -0,0 +1,124 @@
...
+      - key: '.metadata.deletionTimestamp'
+        operator: 'DoesNotExist'
+      - key: '.status.phase'
+        operator: 'In'
+        values:
+          - 'Running'
+      - key: '.metadata.ownerReferences.[].kind'
+        operator: 'In'
+        values:
+          - 'Job'
```
Huang-Wei, line 86: If I want to simulate a plain Pod (not controlled by Job) to go through created->running->completed, I assume I can plumb particular labels to the Pod and match here, right?

stages/pod-fast.yaml
```diff
@@ -0,0 +1,124 @@
...
+  name: pod-delete
+spec:
+  resourceRef:
+    apiGroup: v1
+    kind: Pod
+  selector:
+    matchExpressions:
+      - key: '.metadata.deletionTimestamp'
+        operator: 'Exists'
+  next:
```
Huang-Wei, line 121: I assume `delay` is a general directive so I can craft one here to simulate terminationSeconds?

pkg/apis/v1alpha1/stage_types.go
```diff
@@ -0,0 +1,163 @@
...
+	APIGroup string `json:"apiGroup,omitempty"`
+	// Kind of the referent.
+	Kind string `json:"kind"`
+}
+
+// StageDelay describes the delay time before going to next.
+type StageDelay struct {
+	// DurationSeconds indicates the stage delay time (in seconds).
+	// If JitterDurationSeconds is less than DurationSeconds, then JitterDurationSeconds is used.
+	DurationSeconds int `json:"durationSeconds"`
```
Huang-Wei, line 68: IMO Seconds -> MilliSeconds would be more fine-grained.  (applies to other XYZSeconds as well)

(the upstream also met this issue due to hardcode to seconds in first place: https://github.com/kubernetes/kubernetes/issues/76951)

pkg/config/config.go
```diff
@@ -123,6 +123,18 @@ func Load(ctx context.Context, path ...string) ([]metav1.Object, error) {
 				return nil, err
 			}
 			kwokctlConfiguration = out
+		case v1alpha1.StageKind:
```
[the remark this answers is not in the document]
wzshiming: Yes, `Stage` is a configuration option that remains constant during kwok controller startup and could be made into a CRD in the future.

stages/pod-fast.yaml
wzshiming: Yeah.

stages/pod-fast.yaml
wzshiming: Yeah, `delay` is a generic option that can be added to any `Stage`, but `pod-fast` and node-fast are compatible with the original logic, so there is no `delay`.


Something like this can be done by taking a random value between 1 second and `.metadata.deletionTimestamp` as the delay
``` yaml
  delay:
    durationSeconds: 1
    jitterDurationSecondsFrom:
      expressionFrom: '.metadata.deletionTimestamp'
```

pkg/apis/v1alpha1/stage_types.go
wzshiming: I don't have a strong opinion on this issue.
If we use `MilliSeconds`, should we add both `Seconds` and `MilliSeconds` as they do upstream?

In my opinion, maybe it's more generic if we just use `time.Duration`?

``` go
Duration time.Duration `json:"duration"`
```

``` yaml
delay:
  duration: 1s
```

site/content/en/docs/user/stages-configuration.md
```diff
@@ -0,0 +1,105 @@
...
+    matchExpressions:
+      - key: <expressions-string>
+        operator: <string>
+        values:
+          - <string>
+  delay:
+    durationSeconds: <int>
+    durationSecondsFrom: <expressions-string>
+    jitterDurationSeconds: <int>
+    jitterDurationSecondsFrom: <expressions-string>
```
wzshiming, line 39:
```suggestion
  delay:
    duration: <duration-string>
    durationFrom: <expressions-string>
    jitterDuration: <duration-string>
    jitterDurationFrom: <expressions-string>
```

pkg/apis/v1alpha1/stage_types.go
Huang-Wei: > should we add both Seconds and MilliSeconds as they do upstream?

Nope, just MilliSeconds. The upstream adds both for backward compatibility.

site/content/en/docs/user/stages-configuration.md
Huang-Wei: The old format was right - [api-conventions#units](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units)

pkg/apis/v1alpha1/stage_types.go
Huang-Wei: > In my opinion, maybe it's more generic if we just use time.Duration?

Practically, it's _ok_ if we stick it to a config option. However, if we want better extensibility (like implement in CRD) and non-Go support, we'd better follow the abovementioned API conventions, in particular:

> Duration fields must be represented as integer fields with units being part of the field name (e.g. leaseDurationSeconds). We don't use Duration in the API since that would require clients to implement go-compatible parsing.

wzshiming: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units

I get it, it sounds very reasonable, I will add it in the future.

wzshiming: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units

PR #79, opened 2022-11-23
support pod template change after some duration(for job)

nkuacac:
A way to make pod change status to finish after some durations( defined in the annotation 'last')


wzshiming:
It seems to me that it is not appropriate for a Pod created by a Job to be a direct completer of this behavior.
In fact, the Pod needs to become running and then become completed after some time.

Therefore, we will design an API to try to cover this case, which is still under WIP, feel free to discuss #67.

PR #78, opened 2022-11-22
respect PodCIDR of node

carlory:
Fixes #17 

/kind feature

PR #85, opened 2022-11-25
refactor template cache

carlory:
#### What type of PR is this?

/kind feature

It solves the problem of this parallel test. 

 #78

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kwok/controllers/renderer.go
+++ b/pkg/kwok/controllers/renderer.go
@@ -1,0 +1,89 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"sync"
+	"text/template"
+
+	"sigs.k8s.io/yaml"
+)
+
+type renderer struct {
+	cache      sync.Map
+	bufferPool sync.Pool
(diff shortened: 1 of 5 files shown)


pkg/kwok/controllers/template_cache.go
```diff
@@ -0,0 +1,89 @@
...
+	funcMap    template.FuncMap
+}
+
+func newTemplateCache(funcMap template.FuncMap) *templateCache {
+	return &templateCache{
+		funcMap: funcMap,
+		cache:   sync.Map{},
+		bufferPool: sync.Pool{
+			New: func() interface{} {
+				return new(bytes.Buffer)
```
wzshiming, line 42:
```suggestion
				return bytes.NewBuffer(make([]byte, 4*1024))
```

Maybe it's better to allocate a little memory in advance

pkg/kwok/controllers/template_cache.go
```diff
@@ -0,0 +1,89 @@
...
+			},
+		},
+	}
+}
+
+// renderToJson renders the template with the given text and original object.
+func (tc *templateCache) renderToJson(text string, original interface{}) ([]byte, error) {
```
wzshiming, line 49: Can we add a unit test for this method

pkg/kwok/controllers/template_cache.go
```diff
@@ -0,0 +1,89 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+	"sync"
+	"text/template"
+
+	"sigs.k8s.io/yaml"
+)
+
+type templateCache struct {
```
wzshiming, line 30: I think it would be better to call it Renderer

carlory:
cc @wzshiming

pkg/kwok/controllers/renderer.go
```diff
@@ -0,0 +1,89 @@
...
+		bufferPool: sync.Pool{
+			New: func() interface{} {
+				return bytes.NewBuffer(make([]byte, 4*1024))
+			},
+		},
+	}
+}
+
+// renderToJson renders the template with the given text and original object.
+func (tc *renderer) renderToJson(text string, original interface{}) ([]byte, error) {
```
wzshiming, line 49:
```suggestion
func (r *renderer) renderToJson(text string, original interface{}) ([]byte, error) {
```

Issue #107, opened 2022-11-30
[UMBRELLA] Code Cleanup


wzshiming:
``` bash
cat .golangci.yaml | grep '# - ' | awk '{print $3}'
```

https://golangci-lint.run/usage/linters

- [x] errcheck #118
- [x] errorlint #109
- [x] gocritic #112
- [x] gocyclo #173
- [x] gosec #172
- [x] gosimple #111
- [x] importas #111
- [x] nilerr #110
- [x] prealloc #110
- [x] revive #234
- [ ] rowserrcheck [ignore]
- [x] staticcheck #145
- [x] stylecheck #136
- [x] typecheck #108
- [x] usestdlibvars #114
- [x] whitespace #108

### Install golangci-lint

```
go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.50.1
```

### Uncomment/Add on the checks you want to cleanup

https://github.com/kubernetes-sigs/kwok/blob/main/.golangci.yaml#L3

### Check

Run `golangci-lint run -c .golangci.yaml` based on root of this repo

### Clean up

Clean up the code based on the error reported by the check

## Example

#108

Zhuzhenghao:
I will clean up : 
 rowserrcheck
 staticcheck
 stylecheck

carlory:
I will clean up :

staticcheck
stylecheck

Songjoy:
I will clean up :
errcheck
revive
rowserrcheck

kinzhi:
I will clean up :
 gocyclo
 gosec

kinzhi:
I will clean up :

- gocyclo
- gosec

my-git9:
I will clean up :
 - revive
-  gocritic

wlp1153468871:
I will clean up :

staticcheck
stylecheck

Songjoy:
@wzshiming can you give me some guide ?

```
➜  kwok git:(rowserrcheck) ✗ go version
go version go1.19.2 darwin/arm64

➜  kwok git:(rowserrcheck) ✗ golangci-lint run -c .golangci.yaml
WARN [linters_context] rowserrcheck is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649.
```

wzshiming:
> @wzshiming can you give me some guide ?
> 
> ```
> ➜  kwok git:(rowserrcheck) ✗ go version
> go version go1.19.2 darwin/arm64
[...]

PR #110, opened 2022-11-30
Cleanup for nilerr and  prealloc

Zhuzhenghao:
Signed-off-by: zhuzhenghao <zhenghao.zhu@daocloud.io>

#### What type of PR is this?
/kind cleanup

Fixes #
xref  https://github.com/kubernetes-sigs/kwok/issues/107

```release-note
NONE
```

```docs

```


pkg/kwokctl/utils/cmd.go
```diff
@@ -101,7 +101,7 @@ func ForkExecRestart(ctx context.Context, dir string, name string) error {
 func ForkExecKill(ctx context.Context, dir string, name string) error {
 	pidPath := PathJoin(dir, "pids", filepath.Base(name)+".pid")
 	if _, err := os.Stat(pidPath); err != nil {
-		return nil
+		return err
```
wzshiming, line 104:
```suggestion
	_, err := os.Stat(pidPath); 
	if err != nil {
		// No pid file exists, which means the process has been terminated
		return nil
```

wzshiming: Maybe add a Debug log.

wzshiming (changes requested):

pkg/kwokctl/utils/cmd.go
Zhuzhenghao: > Maybe add a Debug log.

okk, I will add a Debug log

pkg/kwokctl/utils/cmd.go
```diff
@@ -100,7 +100,10 @@ func ForkExecRestart(ctx context.Context, dir string, name string) error {
 
 func ForkExecKill(ctx context.Context, dir string, name string) error {
 	pidPath := PathJoin(dir, "pids", filepath.Base(name)+".pid")
-	if _, err := os.Stat(pidPath); err != nil {
+	_, err := os.Stat(pidPath)
+	if err != nil {
+		fmt.Printf("get fileInfo of pid file %s: %v", pidPath, err)
+		// No pid file exists, which means the process has been terminated
```
wzshiming, line 106:
```suggestion
		// No pid file exists, which means the process has been terminated
		logger := log.FromContext(ctx)
		logger = logger.Debug("Stat file", 
			"path", pidPath,
			"err", err,
		)
```

wzshiming (changes requested):

pkg/kwokctl/utils/cmd.go
Zhuzhenghao: Changed as expected. Thanks for the advice 😁

PR #124, opened 2022-12-02
Fix the fetch platform error

Fish-pro:
#### What type of PR is this?

/kind bug

When executing make build on macos, the following error occurs and the binary target location is incorrect. This pr fixes this error
```sh
➜  kwok git:(main) ✗ make build
./hack/get-platform.sh: line 45: ${os,,}: bad substitution
GOOS= GOARCH=amd64 go build -ldflags '-X sigs.k8s.io/kwok/pkg/consts.Version=v1.0-2-gc70ae71 -X sigs.k8s.io/kwok/pkg/consts.KubeVersion=v1.25.3' -o ./bin//amd64/kwok ./cmd/kwok
GOOS= GOARCH=amd64 go build -ldflags '-X sigs.k8s.io/kwok/pkg/consts.Version=v1.0-2-gc70ae71 -X sigs.k8s.io/kwok/pkg/consts.KubeVersion=v1.25.3' -o ./bin//amd64/kwokctl ./cmd/kwokctl

```

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/hack/get-platform.sh
+++ b/hack/get-platform.sh
@@ -42,7 +42,7 @@ function os() {
   local os
   os="$(uname -s)"
 
-  echo "${os,,}"
+  echo "${os}" | tr '[:upper:]' '[:lower:]'
 }
 
 echo "$(os)/$(arch)"


hack/get-platform.sh
```diff
@@ -42,7 +42,7 @@ function os() {
   local os
   os="$(uname -s)"
 
-  echo "${os,,}"
+  echo "${os}"
```
wzshiming, line 45: This is converting the string to lower case

wzshiming (changes requested):

hack/get-platform.sh
Fish-pro: @wzshiming 
I get it，but error on macos, modified to `echo "${OS}" | tr '[: upper:]' '[: lower:]` please?

Fish-pro: Please remove the `/kind bug` label, or I will close the pr if you don't think it needs fixing。

wzshiming: I checked this syntax was added to bash 4 in 2008, which is 14 years ago.

Perhaps we could add a check for the version of the dependency and output WARNNING if it doesn't match, prompting the user to upgrade the dependency.

Fish-pro: Ok, thanks for your reply

PR #130, opened 2022-12-02
Delete unnecessary and useless judgments

Fish-pro:
#### What type of PR is this?

/kind cleanup

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kwokctl/cmd/etcdctl/etcdctl.go
+++ b/pkg/kwokctl/cmd/etcdctl/etcdctl.go
@@ -66,14 +66,9 @@ func runE(ctx context.Context, flags *flagpole, args []string) error {
 		return err
 	}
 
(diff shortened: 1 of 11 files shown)


wzshiming (commented):
Personally, I think it's just a matter of code style, and both styles have their benefits.

Personally, I think it's just a matter of code style, and both styles have their benefits, so I think both styles should be allowed.

It would be great if we could find a linter to help with the restrictions on styles.

https://golangci-lint.run/usage/linters/

Fish-pro:
> 
> Personally, I think it's just a matter of code style, and both styles have their benefits, so I think both styles should be allowed.
> 
> It would be great if we could find a linter to help with the restrictions on styles.
> 
> https://golangci-lint.run/usage/linters/

Okay, I'm closing it then

PR #142, opened 2022-12-07
check DeletionTimestamp when pod controller receives an add event

Garrybest:
Signed-off-by: Garrybest <garrybest@foxmail.com>

#### What type of PR is this?
/kind bug

When the pod controller receives an add event, it could be that the pod is created, or the kwok-controller is restarted. So if there are a lot of pods in cluster which need to be deleted and the kwok-controller is restarted in coincidence, the rest pods will not be deleted any more.

Fixes #

```release-note
NONE
```

```docs

```


wzshiming (approved):
/approve
/lgtm

PR #146, opened 2022-12-08
Expand home dir before using kubeconfig for kwok

zwpaper:
Signed-off-by: zwPapEr <kweizh@gmail.com>

#### What type of PR is this?

/kind bug

`kwok` can not recognise `--kubeconfig=~/someconfig`, this PR is expanding the homedir `~` to make it work

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kwok/cmd/root.go
+++ b/pkg/kwok/cmd/root.go
@@ -41,6 +41,7 @@ import (
 	"sigs.k8s.io/kwok/pkg/kwok/controllers/templates"
 	"sigs.k8s.io/kwok/pkg/log"
 	"sigs.k8s.io/kwok/pkg/utils/envs"
+	"sigs.k8s.io/kwok/pkg/utils/path"
 )
 
 type flagpole struct {
@@ -65,6 +66,7 @@ func NewCommand(ctx context.Context) *cobra.Command {
 		Version:       consts.Version,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if flags.Kubeconfig != "" {
+				flags.Kubeconfig = path.ExpandHome(flags.Kubeconfig)
 				f, err := os.Stat(flags.Kubeconfig)
 				if err != nil || f.IsDir() {
 					flags.Kubeconfig = ""
--- a/pkg/utils/path/path.go
+++ b/pkg/utils/path/path.go
@@ -1,0 +1,42 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package path
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// ExpandHome expands home directory in file paths.
+func ExpandHome(path string) string {
+	if len(path) == 0 {
+		return path
+	}
+	if path[0] != '~' {
+		return path
+	}
+	if len(path) > 1 && path[1] != '/' && path[1] != '\\' {
+		return path
+	}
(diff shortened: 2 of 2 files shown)


pkg/kwok/cmd/root.go
```diff
@@ -79,6 +80,10 @@ func NewCommand() *cobra.Command {
 		Version:       consts.Version,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if kubeconfig != "" {
+				expanded, err := homedir.Expand(kubeconfig)
```
wzshiming, line 83: Use `os.UserHomeDir()` of the standard library to get the home dir, then replace the prefix ~

wzshiming (changes requested):

pkg/kwok/cmd/root.go
zwpaper: use `os.UserHomeDir()` to replace `~` can definitely fix this simple case, but why not use light-weight lib `homedir` to handle it better?

wzshiming: The main point is that there is no particularly strong reason to look for the user's home directory, and if it is not available, just respond to the error and let the user decide for themselves.

zwpaper: it is not reasonable to force users to use the args without the `=`, most of the docs, yaml configurations use the `=` by default.

we are already looking for users' home directory if we use `os.UserHomeDir()`, so doing it better by using the `homedir` seems more reasonable to me.

but of cause, we can use the simple `os.UserHomeDir` replacement if you insist.

wzshiming: Cool, I want to wrap the `os.UserHomeDir()` into a separate utils function that wraps each place that uses the local path.

zwpaper:
@wzshiming updated, PTAL

pkg/utils/path/path.go
```diff
@@ -0,0 +1,37 @@
...
+	"strings"
+)
+
+// ExpandHome replace the `~` prefix in path, and then return it,
+// Will return the origin one if no `~` prefix found.
+func ExpandHome(path string) string {
+	if len(path) == 0 || path[0] != '~' {
+		return path
+	}
+
```
wzshiming, line 30: if path[1] != '/' then return path

pkg/utils/path/path.go
```diff
@@ -0,0 +1,37 @@
...
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return path
+	}
+
+	return strings.Replace(path, "~", home, 1)
```
wzshiming, line 36: return filepath.Join(home, path[1:])

wzshiming (changes requested):

pkg/utils/path/path.go
```diff
@@ -0,0 +1,39 @@
...
+	"os"
+	"path/filepath"
+)
+
+// ExpandHome replace the `~` prefix in path, and then return it,
+// Will return the origin one if no `~` prefix found.
+func ExpandHome(path string) string {
+	if len(path) == 0 ||
+		path[0] != '~' ||
+		(len(path) > 1 && path[1] != '/') {
```
wzshiming, line 29: `\` is also a legal delimiter on Windows, so we should handle that as well

pkg/utils/path/path.go
```diff
@@ -0,0 +1,39 @@
...
+		return path
+	}
```
wzshiming, line 31:
```suggestion
	if len(path) == 0 {
		return path
	}

	if path[0] != '~' {
		return path
	}

	if len(path) > 1 && path[1] != '/' && path[1] != '\\' {
		return path
	}

```

pkg/utils/path/path.go
```diff
@@ -0,0 +1,39 @@
...
+
+package path
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// ExpandHome replace the `~` prefix in path, and then return it,
+// Will return the origin one if no `~` prefix found.
```
wzshiming, line 25:
```suggestion
// ExpandHome expands home directory in file paths.
```

zwpaper:
Ooops, I have checked the commit history before, and found that the are plenty of PR merged commits, I  thought that squash merged is enabled in PRs...

Updated.

wzshiming:
NVM, appreciates your contributions

/approve
/lgtm

PR #155, opened 2022-12-12
Fix the error log output empty problem

Fish-pro:
#### What type of PR is this?

/kind bug

If an error is generated, the following output is printed. The reason is that when `any` type is converted to `string`, the `json` package `string` becomes empty after it is used. This pr solves this problem
```sh
➜  kwok kwokctl adfadfd
ERROR Execute exit                                                                              err={}
```

Fixes #

```release-note
Fix the error log output empty problem
```

```docs

```

### DIFF
--- a/pkg/log/logger_ctl.go
+++ b/pkg/log/logger_ctl.go
@@ -56,11 +56,18 @@ func formatValue(val slog.Value) string {
 	case slog.StringKind:
 		return quoteIfNeed(val.String())
 	default:
-		v, err := json.Marshal(val.Any())
-		if err == nil {
-			return string(v)
+		switch x := val.Any().(type) {
+		case error:
+			return quoteIfNeed(x.Error())
+		case fmt.Stringer:
+			return quoteIfNeed(x.String())
+		default:
+			v, err := json.Marshal(x)
+			if err == nil {
+				return string(v)
+			}
+			return quoteIfNeed(val.String())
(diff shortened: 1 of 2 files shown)


pkg/log/logger_ctl.go
```diff
@@ -56,11 +56,18 @@ func formatValue(val slog.Value) string {
...
+		case error:
+			var err error
+			err = x
+			return err.Error()
+		default:
+			v, err := json.Marshal(x)
+			if err == nil {
+				return string(v)
+			}
+			return quoteIfNeed(val.String())
```
wzshiming, line 69:
```suggestion
		switch x := val.Any().(type) {
		case error:
			return quoteIfNeed(x.Error())
		case fmt.Stringer:
			return quoteIfNeed(x.String())
		default:
			v, err := json.Marshal(x)
			if err == nil {
				return string(v)
			}
			return quoteIfNeed(val.String())
		}
```

wzshiming (approved):
/approve
/lgtm

wzshiming:
``` console
error validating "pkg/log/logger_ctl_test.go": the file is missing a boilerplate
```

Fish-pro:
> ```
> error validating "pkg/log/logger_ctl_test.go": the file is missing a boilerplate
> ```

@wzshiming Thanks. Got it，updated

PR #166, opened 2022-12-16
setKwokctlConfigurationDefaults function refactor

qingwave:
Signed-off-by: qingwave <isguory@gmail.com>

#### What type of PR is this?
/kind cleanup

Reduce cyclomatic complexity of function [setKwokctlConfigurationDefaults](https://github.com/kubernetes-sigs/kwok/blob/739c461f4b0218f57a26b9793f13f6e3a0ea7ff1/pkg/config/vars.go#L109)

Part of https://github.com/kubernetes-sigs/kwok/issues/107

```release-note
None
```

```docs

```

### DIFF
--- a/pkg/config/vars.go
+++ b/pkg/config/vars.go
@@ -122,16 +122,6 @@ func setKwokctlConfigurationDefaults(config *v1alpha1.KwokctlConfiguration) *v1a
 	}
 	conf.KubeVersion = addPrefixV(envs.GetEnvWithPrefix("KUBE_VERSION", conf.KubeVersion))
 
-	if conf.EtcdVersion == "" {
-		conf.EtcdVersion = k8s.GetEtcdVersion(parseRelease(conf.KubeVersion))
-	}
-	conf.EtcdVersion = trimPrefixV(envs.GetEnvWithPrefix("ETCD_VERSION", conf.EtcdVersion))
-
-	if conf.PrometheusVersion == "" {
-		conf.PrometheusVersion = consts.PrometheusVersion
-	}
-	conf.PrometheusVersion = addPrefixV(envs.GetEnvWithPrefix("PROMETHEUS_VERSION", conf.PrometheusVersion))
-
 	if conf.SecurePort == nil {
 		conf.SecurePort = ptrVar(parseRelease(conf.KubeVersion) > 19)
 	}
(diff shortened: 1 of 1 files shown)


pkg/config/vars.go
```diff
@@ -117,21 +117,6 @@ func setKwokctlConfigurationDefaults(config *v1alpha1.KwokctlConfiguration) *v1a
 	}
 	conf.KwokVersion = addPrefixV(envs.GetEnvWithPrefix("VERSION", conf.KwokVersion))
 
-	if conf.KubeVersion == "" {
```
wzshiming, line 120: This value will be used frequently in the following and needs to be initialized at the first

qingwave: Updated

pkg/config/vars.go
```diff
@@ -142,6 +132,56 @@ func setKwokctlConfigurationDefaults(config *v1alpha1.KwokctlConfiguration) *v1a
...
+
+	if conf.KwokImagePrefix == "" {
+		conf.KwokImagePrefix = consts.ImagePrefix
+	}
+	conf.KwokImagePrefix = envs.GetEnvWithPrefix("IMAGE_PREFIX", conf.KwokImagePrefix)
+
+	if conf.KwokControllerImage == "" {
+		conf.KwokControllerImage = joinImageURI(conf.KwokImagePrefix, "kwok", conf.KwokVersion)
+	}
+	conf.KwokControllerImage = envs.GetEnvWithPrefix("CONTROLLER_IMAGE", conf.KwokControllerImage)
```
wzshiming, line 171: This part could probably also be abstracted into a separate function to set

pkg/config/vars.go
```diff
@@ -142,6 +132,56 @@ func setKwokctlConfigurationDefaults(config *v1alpha1.KwokctlConfiguration) *v1a
 	}
 	conf.QuietPull = ptrVar(envs.GetEnvBoolWithPrefix("QUIET_PULL", *conf.QuietPull))
 
+	setKwokctlKuberentesConfig(conf)
```
wzshiming, line 135:
```suggestion
```

pkg/config/vars.go
```diff
@@ -142,6 +132,56 @@ func setKwokctlConfigurationDefaults(config *v1alpha1.KwokctlConfiguration) *v1a
...
+	if conf.CacheDir == "" {
+		conf.CacheDir = path.Join(WorkDir, "cache")
+	}
+
+	if conf.BinSuffix == "" {
+		if runtime.GOOS == "windows" {
+			conf.BinSuffix = ".exe"
+		}
+	}
+
```
wzshiming, line 152:
```suggestion

	setKwokctlKuberentesConfig(conf)

```

wzshiming:
I tried github's `Apply suggestions from code review`, and it seems to push directly, ignoring my commit please.

pkg/config/vars.go
```diff
@@ -142,6 +132,57 @@ func setKwokctlConfigurationDefaults(config *v1alpha1.KwokctlConfiguration) *v1a
 	}
 	conf.QuietPull = ptrVar(envs.GetEnvBoolWithPrefix("QUIET_PULL", *conf.QuietPull))
 
+
+	v1alpha1.SetObjectDefaults_KwokctlConfiguration(config)
```
qingwave, line 136: Hi @wzshiming, how about the `v1alpha1.SetObjectDefaults_KwokctlConfiguration`, this function should be called at first, or is there some sepcial order?

wzshiming: `v1alpha1.SetObjectDefaults_KwokctlConfiguration` This function sets the default value for variables that are not affected by other variables, but it may affect other variables, try to put it a little above.

qingwave: > Hi @wzshiming, how about the `v1alpha1.SetObjectDefaults_KwokctlConfiguration`, this function should be called at first, or is there some sepcial order?

Look the code again, we should put it **at last**, only cannot find variables from env or flags and then set the default value via `v1alpha1.SetObjectDefaults_KwokctlConfiguration`, if not, the default value(set by ObjectDefaults) will override users's config. Please correct me if I miss something.

wzshiming: Makes sense

Issue #232, opened 2023-01-25
[UMBRELLA] Add `hack/update-*.sh`


wzshiming:
### What would you like to be added?

- [x] #226
- [ ] update-ends-newline.sh #243
- [x] update-go-format.sh #241
- [x] update-go-mod.sh #242

### Why is this needed?

Optimize the development experience

PR #241, opened 2023-01-29
build: add update-go-format.sh to do go format

Fish-pro:
Signed-off-by: Fish-pro <zechun.chen@daocloud.io>

#### What type of PR is this?

/kind feature

Part of #232

```release-note
NONE
```

```docs

```

### DIFF
--- a/hack/update-go-format.sh
+++ b/hack/update-go-format.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright 2022 The Kubernetes Authors.
+# Copyright 2023 The Kubernetes Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -17,17 +17,8 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-ROOT_DIR="$(dirname "${BASH_SOURCE[0]}")/.."
+REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 
-function check() {
-  echo "Verify gofmt"
-  out="$(gofmt -l -d $(find cmd pkg -name '*.go'))"
-  if [[ -n "${out}" ]]; then
-    echo "${out}"
-    return 1
-  fi
-}
+cd "${REPO_ROOT}"
 
-cd "${ROOT_DIR}"
-
-check || exit 1
+gofmt -s -w $(find cmd pkg -name '*.go')
--- a/hack/update-all.sh
+++ b/hack/update-all.sh
@@ -22,7 +22,13 @@ ROOT_DIR=$(realpath "$(dirname "${BASH_SOURCE[0]}")"/..)
 failed=()
 
 if [[ "${UPDATE_CODEGEN:-true}" == "true" ]]; then
+    echo "[*] Update codegen..."
     "${ROOT_DIR}"/hack/update-codegen.sh || failed+=(codegen)
+fi
+
+if [[ "${UPDATE_GO_FORMAT:-true}" == "true" ]]; then
+    echo "[*] Update go format..."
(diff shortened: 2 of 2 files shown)


hack/update-go-format.sh
```diff
@@ -0,0 +1,37 @@
...
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+source "${REPO_ROOT}"/hack/util.sh
+
+util::verify_go_version
```
wzshiming, line 23:
```suggestion
```

It seems that `go` is not used here, so there should be no need to check if it exists

hack/update-go-format.sh
```diff
@@ -0,0 +1,37 @@
...
+
+cd "${REPO_ROOT}"
+find_files() {
```
wzshiming, line 26:
```suggestion
function find_files() {
```

hack/update-go-format.sh
```diff
@@ -0,0 +1,37 @@
...
+  find . -not \( \
+      \( \
+        -wholename './bin' \
+        -o -wholename './.git' \
+        -o -wholename './site/*' \
+        -o -wholename '*/vendor/*' \
+      \) -prune \
+    \) -name '*.go'
+}
```
wzshiming, line 35: The go files only exist under `cmd` and `pkg`, so maybe we should just find these two directories, and change them together in `verify-go-format.sh`

hack/update-go-format.sh
Fish-pro: I understand that `gofmt` depends on the `go` installation, so it needs to be verified
image

wzshiming: `gofmt` is a standalone binary file and does not depend on `go`

https://github.com/golang/go/blob/master/src/cmd/gofmt/gofmt.go

Fish-pro: Thanks for tips

hack/verify-go-format.sh
```diff
@@ -21,7 +21,7 @@ ROOT_DIR="$(dirname "${BASH_SOURCE[0]}")/.."
 
 function check() {
   echo "Verify gofmt"
-  out="$(gofmt -l -d $(find . -name '*.go'))"
+  out="$(gofmt -l -d $(find pkg -name '*.go'))"
```
wzshiming, line 24: What about `cmd`?

Fish-pro: Sorry, It's not saved. It's local

hack/verify-go-format.sh
```diff
@@ -21,7 +21,12 @@ ROOT_DIR="$(dirname "${BASH_SOURCE[0]}")/.."
 
 function check() {
   echo "Verify gofmt"
-  out="$(gofmt -l -d $(find . -name '*.go'))"
+  out="$(gofmt -l -d $(find cmd -name '*.go'))"
+  if [[ -n "${out}" ]]; then
+    echo "${out}"
+    return 1
+  fi
+  out="$(gofmt -l -d $(find pkg -name '*.go'))"
```
wzshiming, line 29:
```suggestion
  out="$(gofmt -l -d $(find cmd pkg -name '*.go'))"
```

hack/update-go-format.sh
```diff
@@ -0,0 +1,28 @@
...
+
+REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+
+cd "${REPO_ROOT}"
+function find_files() {
+  find cmd -name '*.go'
+  find pkg -name '*.go'
+}
+
+find_files | xargs gofmt -s -w
```
wzshiming, line 28:
```suggestion
gofmt -s -w $(find cmd pkg -name '*.go')
```

wzshiming (approved):
/approve
/lgtm

Issue #224, opened 2023-01-18
proposal: auto detect docker or nerdctl runtime when not specified a runtime flag


zwpaper:
### What would you like to be added?

1. change the runtime flag default to empty
2. auto detect by checking whether the docker/nerdctl executable file existed in PATH

I would like to implement it if accepted.

### Why is this needed?

nerdctl became more popular as a docker replacement, we have to specify the runtime args or add an env variable for nerdctl users.

it would be more convenience if we can detect it by ourself.

wzshiming:
Thanks for the proposal, but I have a few questions about the proposal.
there are currently 4 runtimes, binary, docker, nerdctl and kind, is your idea to check only docker and nerdctl? 

> auto detect by checking whether the docker/nerdctl executable file existed in PATH

If just checking the binary is not reliable, a more reliable way to check if it is working properly is needed.

In my opinion, the binary runtime is a better choice than nerdctl for Linux, because it works in basically any environment, while nerdctl is required to be installed properly. WDYT?

zwpaper:
Hi @wzshiming,

> there are currently 4 runtimes, binary, docker, nerdctl and kind, is your idea to check only docker and nerdctl?

I noticed that kind uses docker to pull images so I am assuming kind could only work when docker is available, it should also work when nerdctl is available by the way.

I recheck the code, and found out the kind executable file is also needed when specify as runtime, so it would also be necessary to check it at the same time.

all in all, kwok should check all 4 runtimes in order when no one is specified by users.
the only question is what should the order be.

> If just checking the binary is not reliable, a more reliable way to check if it is working properly is needed.

agree, checking the binary is the basic, checking if it is working properly would be better,
maybe we can run some basic command like `version`

> In my opinion, the binary runtime is a better choice than nerdctl for Linux, because it works in basically any environment, while nerdctl is required to be installed properly. WDYT?

partially agree, binary runtime is great and simple to use for Linux, but sometimes we want to keep the system clean, I would choose to use a container runtime than run it natively.

what's more, nerdctl is the only choice in macOS except for docker

zwpaper:
by the way, I am using a mac to run kwokctl, and not yet found problems, I would submit a PR to update the docs about the macOS part later.

wzshiming:
> I noticed that kind uses docker to pull images so I am assuming kind could only work when docker is available, it should also work when nerdctl is available by the way.

Kind only supports docker and podman, and does not support nerdctl.

[...]

PR #244, opened 2023-02-01
feat: detect runtime automatically

zwpaper:
Signed-off-by: WeiZhang <kweizh@gmail.com>

#### What type of PR is this?
/kind feature

Fixes https://github.com/kubernetes-sigs/kwok/issues/224#event-8395695133

1. drop default runtime args value `docker`
2. check docker,kind executable, nerdctl minimal version, kind in linux
3. automatically use runtime by order: docker, nerdctl, binary, kind
4. nerdctl v1.2.0 support compose start and stop, used as minimal supported

```release-note
Automatically detect runtime in order: docker, nerdctl, binary
```

```docs

```


zwpaper:
cc @wzshiming

pkg/kwokctl/cmd/create/cluster/cluster.go
```diff
@@ -127,14 +128,35 @@ func runE(ctx context.Context, flags *flagpole) error {
...
 	}
 
-	buildRuntime, ok := runtime.DefaultRegistry.Get(flags.Options.Runtime)
-	if !ok {
-		return fmt.Errorf("runtime %q not found", flags.Options.Runtime)
+	var runtimeCandidate []string
+	if flags.Options.Runtime != "" {
+		runtimeCandidate = []string{flags.Options.Runtime}
+	} else {
+		runtimeCandidate = []string{
```
wzshiming, line 135: Add this to the KwokctlConfiguration to configure

pkg/kwokctl/runtime/binary/cluster.go
```diff
@@ -50,6 +51,10 @@ type Cluster struct {
 
 // NewCluster creates a new Runtime for binary
 func NewCluster(name, workdir string) (runtime.Runtime, error) {
+	if goruntime.GOOS != "linux" {
+		return nil, fmt.Errorf("binary: only supported in linux")
+	}
+
```
wzshiming, line 57: Binary is works in non-Linux

pkg/kwokctl/runtime/compose/cluster.go
```diff
@@ -46,13 +46,37 @@ import (
...
-	}, nil
+		driver:  consts.RuntimeTypeNerdctl,
+	}
+
+	v, err := version.ParseFromBinary(context.Background(), cluster.driver)
+	if err != nil {
+		return nil, err
+	}
+	if v.LT(minNerdctlVersion) {
+		return nil, fmt.Errorf("nerdctl: minimal supported version is %s", &minNerdctlVersion)
```
wzshiming, line 67: Why only support the 1.2.0 released yesterday, this approach is not appropriate.

pkg/kwokctl/cmd/create/cluster/cluster.go
```diff
@@ -127,14 +128,35 @@ func runE(ctx context.Context, flags *flagpole) error {
...
-	rt, err := buildRuntime(name, workdir)
+	var rt runtime.Runtime
+	var err error
+	var errs []error
+	for _, r := range runtimeCandidate {
+		buildRuntime, ok := runtime.DefaultRegistry.Get(r)
+		if !ok {
+			continue
+		}
+		rt, err = buildRuntime(name, workdir)
```
wzshiming, line 151: We need to add a new method to the Cluster interface to check if it can run

wzshiming (changes requested):

pkg/kwokctl/runtime/binary/cluster.go
zwpaper: then how should we detect whether binary works?

pkg/kwokctl/cmd/create/cluster/cluster.go
zwpaper: the `buildRuntime` has returned an error previously so I used it to do the basic runtime checks.

we can discuss a new method if necessary

pkg/kwokctl/runtime/compose/cluster.go
wzshiming: We need to maintain compatibility with versions prior to 1.2.0

pkg/kwokctl/cmd/create/cluster/cluster.go
wzshiming: Yeah, add a new method, not in NewCluster

pkg/kwokctl/runtime/binary/cluster.go
wzshiming: I was thinking that the default order for Linux settings is docker, nerdctl and binary. and for everything else just check docker and nerdctl.

pkg/kwokctl/cmd/create/cluster/cluster.go
zwpaper: Kwok load config in runtime, we may need some refactoring to load the config before automatically selecting a runtime.

pkg/kwokctl/runtime/binary/cluster.go
zwpaper: like the previous one, it needs some refactoring to set the config, how about hardcoded currently?

users can always specify one if they want to

pkg/kwokctl/runtime/compose/cluster.go
zwpaper: reverted changes and drop the limitation

pkg/kwokctl/cmd/create/cluster/cluster.go
zwpaper: may I ask why preferring a new method? what's the point of returning err for `NewCluster` if we always return nil?

pkg/kwokctl/cmd/create/cluster/cluster.go
wzshiming: I think no need to refactor the configuration loading, 

Also, do the runtime detection here based on the configuration.

pkg/kwokctl/cmd/create/cluster/cluster.go
wzshiming: Runtime detection is only required when creating a cluster, and is not required and cannot be switched in other.

pkg/kwokctl/runtime/compose/cluster.go
wzshiming: This work is nearly complete in another PR kubernetes-sigs/kwok#249

pkg/kwokctl/runtime/binary/cluster.go
wzshiming: I don't think hard coding is appropriate.

wzshiming: Set up a default for KwokctlConfiguration https://github.com/kubernetes-sigs/kwok/blob/122f36ce87128371d2b9bbcff43837d764b35e37/pkg/config/vars.go#L123

pkg/apis/v1alpha1/kwokctl_configuration_types.go
```diff
@@ -51,7 +51,6 @@ type KwokctlConfigurationOptions struct {
 
 	// Runtime is the runtime to use.
 	// is the default value for flag --runtime and env KWOK_RUNTIME
-	// +default="docker"
 	Runtime string `json:"runtime,omitempty"`
```
wzshiming, line 55:
```suggestion

	// Runtimes is a list of alternate runtimes. When Runtime is empty,
	// the availability of the runtimes in the list is checked one by one and set to Runtime
	Runtimes []string `json:"runtimes,omitempty"`

```

wzshiming, line 55:
```suggestion

	// Runtimes is a list of alternate runtimes. When Runtime is empty,
	// the availability of the runtimes in the list is checked one by one and set to Runtime
	Runtimes []string `json:"runtimes,omitempty"`

```

pkg/kwokctl/cmd/create/cluster/cluster.go
zwpaper: nope, you need to know which runtime to use when even updating, deleting

pkg/kwokctl/runtime/binary/cluster.go
zwpaper: Oops, sorry for not noticed the config is already existed inside flags, will fix this

pkg/kwokctl/cmd/create/cluster/cluster.go
zwpaper: fixed this time, PTAL @wzshiming

pkg/config/vars.go
```diff
@@ -147,6 +147,16 @@ func setKwokctlConfigurationDefaults(config *v1alpha1.KwokctlConfiguration) *v1a
 	conf.QuietPull = format.Ptr(envs.GetEnvWithPrefix("QUIET_PULL", *conf.QuietPull))
 
 	conf.Runtime = envs.GetEnvWithPrefix("RUNTIME", conf.Runtime)
+	if len(conf.Runtimes) == 0 {
```
wzshiming, line 150:
```suggestion
	if conf.Runtime == "" && len(conf.Runtimes) == 0 {
```

pkg/config/vars.go
```diff
@@ -147,6 +147,16 @@ func setKwokctlConfigurationDefaults(config *v1alpha1.KwokctlConfiguration) *v1a
...
+		conf.Runtimes = []string{
+			consts.RuntimeTypeDocker,
+			consts.RuntimeTypeNerdctl,
+			consts.RuntimeTypeKind,
```
wzshiming, line 154: Kind depends on docker. If docker doesn't work, so does Kind.
```suggestion
```

pkg/kwokctl/cmd/create/cluster/cluster.go
wzshiming: Cluster creation will record the runtime, and the second time will use the previous Runtime, instead of re-detecting it.

pkg/kwokctl/runtime/compose/cluster.go
```diff
@@ -46,15 +46,32 @@ import (
...
+		driver:  consts.RuntimeTypeDocker,
+	}, nil
+}
+
+// Available  checks whether the runtime is available.
+func (c *Cluster) Available(ctx context.Context) error {
+	return exec.Exec(context.Background(), "", exec.IOStreams{},
+		c.driver, "version")
+}
+
```
wzshiming, line 74:
```suggestion
// NewCluster creates a new Runtime for docker-compose.
func NewCluster(name, workdir string) (runtime.Runtime, error) {
	return &Cluster{
		Cluster: runtime.NewCluster(name, workdir),
	}, nil
}

// Available  checks whether the runtime is available.
func (c *Cluster) Available(ctx context.Context) error {
	conf, err := c.Config(ctx)
	if err != nil {
		return err
	}

	return exec.Exec(ctx, "", exec.IOStreams{}, config.Options.Runtime, "version")
}

```

Revert other changes in the file

pkg/kwokctl/runtime/compose/init.go
```diff
@@ -22,6 +22,6 @@ import (
 )
 
 func init() {
-	runtime.DefaultRegistry.Register(consts.RuntimeTypeDocker, NewCluster)
-	runtime.DefaultRegistry.Register(consts.RuntimeTypeNerdctl, NewCluster)
+	runtime.DefaultRegistry.Register(consts.RuntimeTypeDocker, NewDockerCluster)
+	runtime.DefaultRegistry.Register(consts.RuntimeTypeNerdctl, NewNerdctlCluster)
```
wzshiming, line 26:
```suggestion
	runtime.DefaultRegistry.Register(consts.RuntimeTypeDocker, NewCluster)
	runtime.DefaultRegistry.Register(consts.RuntimeTypeNerdctl, NewCluster)
```

pkg/kwokctl/runtime/kind/cluster.go
```diff
@@ -52,6 +53,20 @@ func NewCluster(name, workdir string) (runtime.Runtime, error) {
 	}, nil
 }
 
+// Available  checks whether the runtime is available.
+func (c *Cluster) Available(ctx context.Context) error {
+	err := exec.Exec(context.Background(), "", exec.IOStreams{}, consts.RuntimeTypeKind)
```
wzshiming, line 58:
```suggestion
	err := exec.Exec(ctx, "", exec.IOStreams{}, consts.RuntimeTypeKind)
```

pkg/kwokctl/runtime/kind/cluster.go
```diff
@@ -52,6 +53,20 @@ func NewCluster(name, workdir string) (runtime.Runtime, error) {
...
+	if err != nil {
+		return err
+	}
+
+	// kind depends on docker
+	// TODO: nerdctl kind provider support is discussing in
+	// https://github.com/kubernetes-sigs/kind/issues/2317 and
+	// https://github.com/containerd/nerdctl/issues/349
+	return exec.Exec(context.Background(), "", exec.IOStreams{}, "docker")
```
wzshiming, line 67:
```suggestion
	return exec.Exec(ctx, "", exec.IOStreams{}, "docker")
```

wzshiming (changes requested):

pkg/kwokctl/runtime/compose/cluster.go
zwpaper: `cluster.Available` is used to detect whether the runtime is available, at this time, `config.Options.Runtime` is empty, it does not know it is `docker` or `nerdctl`

zwpaper: we exposed `docker` and `nerdctl` as 2 runtimes, IMHO, It is reasonable to offer `NewDockerCluster` and `NewNerdctlCluster`, those 2 runtimes can know what they are and reuse this `compose` logic in most places.

wzshiming: > `cluster.Available` is used to detect whether the runtime is available, at this time, `config.Options.Runtime` is empty, it does not know it is `docker` or `nerdctl`

After creating the cluster, the Runtime is already set and cannot be empty here.

> we exposed `docker` and `nerdctl` as 2 runtimes, IMHO, It is reasonable to offer `NewDockerCluster` and `NewNerdctlCluster`, those 2 runtimes can know what they are and reuse this `compose` logic in most places.

Maybe if we add a parameter runtime to NewCluster, this should be more appropriate and we can leave it unchanged in the future when we add podman-compose. WDYT?

pkg/kwokctl/runtime/kind/cluster.go
wzshiming: Currently Kind actually supports podman, so it's not really a good idea to check docker.

pkg/kwokctl/runtime/compose/cluster.go
zwpaper: > After creating the cluster, the Runtime is already set and cannot be empty here.

check the logic you suggested https://github.com/kubernetes-sigs/kwok/pull/244#discussion_r1095454530, the runtime read from config is set AFTER `rt.Available()`, we will always read an empty runtime from config in `rt.Available()` when we trying to detect runtime automatically.

the reason I added a `driver` param to the compost Cluster is to set it after creating the cluster

> Maybe if we add a parameter runtime to NewCluster
NewCluster only accepts `name` and `workdir`, it is odd if we add a `runtime-kind` parameter to `NewCluster` when we already specified we are creating some kind of runtime like `compose.NewCluster(name, workdir, docker)`, `kind.NewCluster(name, workdir, kind)`

providing `NewNerdctlCluster` and `NewDockerCluster` and registering it to the registry seems more elegant.

> this should be more appropriate and we can leave it unchanged in the future when we add podman-compose.

this is not that possible we can add `podman` support without adding some incompatibility like currently existed `if runtime == Nerdctl`,

we only have to provide a function `NewPodmanCluster` when Kwok supports Podman if we follow the `NewNerdctlCluster`, `NewDockerCluster` way, and maintain the incompatibility inside compose package.

pkg/kwokctl/runtime/kind/cluster.go
zwpaper: I am not that familiar with kind, maybe we can check docker and podman, it is ok when one of them works?

or just leave it returning nil directly?

wzshiming: Yeah, just check Kind.

pkg/kwokctl/runtime/compose/cluster.go
wzshiming: That's fine, but this part of the refactoring is not necessary in this feature, so maybe we can discuss this in the new PR

pkg/kwokctl/runtime/kind/cluster.go
zwpaper: but kind would be downloaded by kwokctl if not existed...

pkg/kwokctl/runtime/compose/cluster.go
zwpaper: I can raise another PR to discuss `NewCluster`, but the there is a problem need to be solved, let me paste a screenshot to explain myself:

image

we can refactor it like this, but it seemed odd

image

image


that is mainly why I make the change

pkg/kwokctl/runtime/kind/cluster.go
zwpaper: now I check docker or podman, because you mentioned kwokctl will download kind if not found, WDYT?

wzshiming: Yeah, You're right, do as you say.

pkg/kwokctl/runtime/compose/cluster.go
wzshiming: Yeah, You're right, do as you say.

pkg/kwokctl/runtime/compose/cluster.go
```diff
@@ -46,15 +46,32 @@ import (
 // Cluster is a implementation of Runtime for docker-compose.
 type Cluster struct {
 	*runtime.Cluster
+
+	driver string
```
wzshiming, line 50:
```suggestion
	runtime string
```

Maintain contextually consistent naming

pkg/kwokctl/runtime/compose/cluster.go
```diff
@@ -46,15 +46,32 @@ import (
...
 	return &Cluster{
 		Cluster: runtime.NewCluster(name, workdir),
+		driver:  consts.RuntimeTypeDocker,
 	}, nil
 }
 
+// Available  checks whether the runtime is available.
+func (c *Cluster) Available(ctx context.Context) error {
+	return exec.Exec(context.Background(), "", exec.IOStreams{},
+		c.driver, "version")
```
wzshiming, line 72:
```suggestion
	return exec.Exec(ctx, "", exec.IOStreams{}, c.runtime, "version")
```

pkg/kwokctl/runtime/kind/cluster.go
```diff
@@ -52,6 +52,25 @@ func NewCluster(name, workdir string) (runtime.Runtime, error) {
...
 }
 
+// Available  checks whether the runtime is available.
+func (c *Cluster) Available(ctx context.Context) error {
+	// kind depends on docker or podman
+	// kwokctl will download kind binary if it is not available
+	// TODO: nerdctl kind provider support is discussing in
+	// https://github.com/kubernetes-sigs/kind/issues/2317 and
+	// https://github.com/containerd/nerdctl/issues/349
+	err := exec.Exec(ctx, "", exec.IOStreams{}, "docker")
```
wzshiming, line 62:
```suggestion
	err := exec.Exec(ctx, "", exec.IOStreams{}, "docker", "version")
```

pkg/kwokctl/runtime/kind/cluster.go
```diff
@@ -52,6 +52,25 @@ func NewCluster(name, workdir string) (runtime.Runtime, error) {
...
+	if err == nil {
+		return nil
+	}
+	err = exec.Exec(ctx, "", exec.IOStreams{}, "podman")
```
wzshiming, line 66:
```suggestion
	err = exec.Exec(ctx, "", exec.IOStreams{}, "podman", "version")
```

pkg/kwokctl/runtime/compose/cluster.go
```diff
@@ -623,6 +640,7 @@ func (c *Cluster) Stop(ctx context.Context) error {
 	subcommand := "stop"
 	if isOldNerdctl {
 		subcommand = "down"
+
```
wzshiming, line 643:
```suggestion
```

wzshiming:
Cool, With this PR basically in place, we can now consider how to add e2e to ensure that this feature is not accidentally broken, which is after all the default boot way of the future.

pkg/kwokctl/runtime/kind/cluster.go
```diff
@@ -52,6 +52,25 @@ func NewCluster(name, workdir string) (runtime.Runtime, error) {
...
+	if err == nil {
+		return nil
+
+	}
+	return fmt.Errorf("docker or podman is required for kind")
```
wzshiming, line 71: I thought about this or not to check the more appropriate

```suggestion
	return nil
```

zwpaper:
image

CI error because restart error, let me dig deeper into it and check whether is related to this changes

zwpaper:
weird, I did not changed stop related part, why the restart testing keep failing

wzshiming:
> weird, I did not changed stop related part, why the restart testing keep failing

This is a flaking test and I'm working on fixing it #257

zwpaper:
@wzshiming I found the problems, and will raise a PR to fix it.

it should be not related to https://github.com/kubernetes-sigs/kwok/pull/257

wzshiming:
I don't know why, but now it seems to be working again, and nothing has changed.

You remove the revet commit and try again.

zwpaper:
@wzshiming Nope, check https://github.com/kubernetes-sigs/kwok/pull/260#issuecomment-1420232954, I updated comments, and tested nerdctl locally, and there are bugs in nerdctl v1.2.0

pkg/kwokctl/cmd/create/cluster/cluster.go
```diff
@@ -127,15 +127,42 @@ func runE(ctx context.Context, flags *flagpole) error {
...
+		buildRuntime, ok := runtime.DefaultRegistry.Get(flags.Options.Runtime)
+		if !ok {
+			return fmt.Errorf("runtime %q not found", flags.Options.Runtime)
+		}
+		rt, err = buildRuntime(name, workdir)
+		if err != nil {
+			return fmt.Errorf("runtime %v not available: %w", flags.Options.Runtime, err)
+		}
 	}
+	logger = logger.With("runtime", flags.Options.Runtime)
```
wzshiming, line 165:
```suggestion
```

pkg/kwokctl/cmd/create/cluster/cluster.go
```diff
@@ -127,15 +127,42 @@ func runE(ctx context.Context, flags *flagpole) error {
...
+			if err = rt.Available(ctx); err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			flags.Options.Runtime = r
+			break
+		}
+		if flags.Options.Runtime == "" {
+			return fmt.Errorf("runtime %v not available: %v", flags.Options.Runtimes, errs)
+		}
```
wzshiming, line 154:
```suggestion
		}
		logger.Debug("Detected runtime available", "runtime", flags.Options.Runtime)
```

wzshiming (commented):
Can we have a test to cover this change

pkg/kwokctl/cmd/create/cluster/cluster.go
zwpaper: as we now detect runtime automatically, it maybe worth logging the runtime?

zwpaper:
> Can we have a test to cover this change

will raise a PR right after this one get merged

wzshiming:
Ah, No, Submitting tests and features separately is not a good way to go in my opinion.

zwpaper:
then we have to keep rebasing, squashing...

but I also understand why you prefer adding the test in one PR, let me have a try later free

.github/workflows/test.yaml
```diff
@@ -105,14 +105,13 @@ jobs:
           - docker
           - nerdctl
           - kind
+          - auto # auto runtime: will detect runtime automatically
```
wzshiming, line 108:
``` yaml
kind: KwokctlConfiguration
options:
  runtimes: ["nerdctl", "docker"]
```

We use this configuration to switch the order, so we don't need to change these.

pkg/config/vars.go
```diff
@@ -147,6 +147,15 @@ func setKwokctlConfigurationDefaults(config *v1alpha1.KwokctlConfiguration) *v1a
 	conf.QuietPull = format.Ptr(envs.GetEnvWithPrefix("QUIET_PULL", *conf.QuietPull))
 
 	conf.Runtime = envs.GetEnvWithPrefix("RUNTIME", conf.Runtime)
+	if len(conf.Runtimes) == 0 {
+		conf.Runtimes = []string{
+			consts.RuntimeTypeDocker,
+			consts.RuntimeTypeNerdctl,
```
wzshiming, line 153: #255 tried to add e2e for nerdctl under darwin, but it doesn't seem to work, so better to just add it for Linux?

zwpaper: previously I was trying to add nerdctl for mac, because I thought it would be suitable for the runtime detection, but your idea using configure file seems easier, would use this way for runtime detection 

Will raise another PR for mac nerdctl, I am a mac nerdctl user, it should be easy to add it

test/kwokctl/kwokctl_binary.test.sh
```diff
@@ -20,9 +20,12 @@ DIR="$(realpath "${DIR}")"
 source "${DIR}/helper.sh"
 
 function main() {
```
wzshiming, line 22: Add a new test file `kwokctl_auto_detect.test.sh`

``` bash
DIR="$(dirname "${BASH_SOURCE[0]}")"

DIR="$(realpath "${DIR}")"

source "${DIR}/helper.sh"

function main() {
  kwokctl create cluster --config "${DIR}"/kwokctl-config-runtimes.yaml || exit 1
  kwokctl delete cluster || exit 1
}

requirements_for_binary

main
```

wzshiming (changes requested):

test/kwokctl/kwokctl_binary.test.sh
wzshiming: Add in `.github/workflows/test.yaml`

``` yaml
  test-kwokctl-auto-detect:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - name: Set up Go
        uses: actions/setup-go@v3
        with:
          go-version: 1.19

      - name: Make pki directory
        if: ${{ matrix.kwokctl-runtime == 'binary' }}
        shell: bash
        run: |
          sudo mkdir -p /var/run/kubernetes
          sudo chmod 777 /var/run/kubernetes

      - name: Test Auto Detect
        shell: bash
        run: |
          ./hack/e2e-test.sh kwokctl/kwokctl_auto_detect

```

zwpaper:
@wzshiming this should be finally good to go, please help to restart the ubuntu nerdctl failed test and check whether we are good to go

wzshiming (commented):
Looks like it's almost ready!

test/kwokctl/kwokctl_auto_detect.test.sh
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+DIR=$(realpath $(dirname "${BASH_SOURCE[0]}"))
+
+source "${DIR}/helper.sh"
+
+function main() {
+  local all_releases=("${@}")
+  for release in "${all_releases[@]}"; do
+    KWOK_KUBE_VERSION="${release}" kwokctl create cluster --config "${DIR}"/kwokctl-config-runtimes.yaml || exit 1
```
wzshiming, line 23:
```suggestion
    KWOK_KUBE_VERSION="${release}" kwokctl -v=-4 create cluster --timeout 10m --wait 10m --quiet-pull --config "${DIR}"/kwokctl-config-runtimes.yaml || exit 1
```

.github/workflows/test.yaml
```diff
@@ -222,3 +222,23 @@ jobs:
...
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19
+
+      - name: Make pki directory
+        shell: bash
+        run: |
+          sudo mkdir -p /var/run/kubernetes
+          sudo chmod 777 /var/run/kubernetes
```
wzshiming, line 244:
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -222,3 +222,8 @@ jobs:
         shell: bash
         run: |
           ./hack/e2e-test.sh kwokctl/kwokctl_${{ matrix.kwokctl-runtime }}_benchmark
+
+      - name: Test Auto Detect
+        shell: bash
```
wzshiming, line 227:
```suggestion
        if: ${{ matrix.kwokctl-runtime == 'binary' }}
        shell: bash
```

pkg/config/vars.go
wzshiming: note: Yeah, Nerdctl as a secondary option, if e2e is not enabled under darwin before the next release, I plan to change it to be used only under Linux, just like the binary.

wzshiming (approved):
Thank you very much for your contribution!

/approve
/lgtm

PR #247, opened 2023-02-02
refactor pkg/kwokctl/cmd/create/cluster/cluster.go

Zhuzhenghao:
#### What type of PR is this?

/kind cleanup

1. Improving the code readability by breaking down the function into smaller sub-functions, using more descriptive variable and function names.

2. Consider breaking down the function into smaller, modular components that can be tested and reused in different parts of the system.

3. Make sure the code is reusable, so it can be used in other parts of the system with little or no modification.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/pkg/kwokctl/cmd/create/cluster/cluster.go
+++ b/pkg/kwokctl/cmd/create/cluster/cluster.go
@@ -114,11 +114,7 @@ func NewCommand(ctx context.Context) *cobra.Command {
 }
 
 func runE(ctx context.Context, flags *flagpole) error {
-	name := config.ClusterName(flags.Name)
-	workdir := path.Join(config.ClustersDir, flags.Name)
-
-	logger := log.FromContext(ctx)
-	logger = logger.With("cluster", flags.Name)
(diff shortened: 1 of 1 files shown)


wzshiming (commented):
I don't think it has reached the limits of [gocyclo](https://github.com/kubernetes-sigs/kwok/blob/122f36ce87128371d2b9bbcff43837d764b35e37/.golangci.yaml#L15). So there is no need to refactor this way, it will lead to a lot of conflicts and no gain, we can focus on the [roadmap](https://github.com/orgs/kubernetes-sigs/projects/21) first.

Zhuzhenghao:
> I don't think it has reached the limits of [gocyclo](https://github.com/kubernetes-sigs/kwok/blob/122f36ce87128371d2b9bbcff43837d764b35e37/.golangci.yaml#L15). So there is no need to refactor this way, it will lead to a lot of conflicts and no gain, we can focus on the [roadmap](https://github.com/orgs/kubernetes-sigs/projects/21) first.

okk

PR #249, opened 2023-02-02
[kwokctl] Update start/stop for nerdctl

Zhuzhenghao:
Signed-off-by: zhuzhenghao <zhenghao.zhu@daocloud.io>

#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

Update start/stop for nerdctl.

1. nerdctl supports `compose start` in [version 1.2.0](https://github.com/containerd/nerdctl/releases/tag/v1.2.0). Add judgement to  determine when not to use `compose start`.
2. same as `compost stop`

#### Which issue(s) this PR fixes:

Fixes #245

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs
NONE
```

PR #260, opened 2023-02-07
Revert "[kwokctl] Update start/stop for nerdctl"

zwpaper:
Reverts kubernetes-sigs/kwok#249

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -160,11 +160,10 @@ jobs:
         if: ${{ matrix.kwokctl-runtime == 'nerdctl' }}
         shell: bash
         run: |
-          curl -sSL https://github.com/containerd/nerdctl/releases/download/v1.2.0/nerdctl-full-1.2.0-linux-amd64.tar.gz -o - | sudo tar -xz -C /usr/local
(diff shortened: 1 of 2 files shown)


wzshiming:
It doesn't look like that's the problem?

wzshiming:
The #258 tests all pass, and it seems that some unstable behavior of nerdctl is the cause.

zwpaper:
it is due to the unstable behavior of `nerdctl stop`:

I build up a cluster:

image

then stop it, it disappears for a while, but reappears seconds later, and the containers up-time keep growing up

image


the test results differ because the difference of containers' disappear time is different each time we call `kwokctl stop cluster`

wzshiming:
Good job.

In fact, the old nerdctl's restart test was also so unstable that I thought we'd add a check to wait for its complete shutdown. wdyt?

zwpaper:
> I thought we'd add a check to wait for its complete shutdown
of cause, will raise a PR after checking the `nerdctl compose stop`

zwpaper:
this is confirmed to a bug, should we let this PR continue? @wzshiming 

https://github.com/containerd/nerdctl/issues/1980#issuecomment-1420715781

wzshiming:
Thanks!

/approve
/lgtm

wzshiming (approved):

PR #299, opened 2023-02-20
Fixes #284: configure hostIP of Pod's status and support hostNetwork

### LINKED ISSUE #284: [kwok] Respect HostNetwork of Pod

If `spec.hostNetwork` is true then `status.podIP` is set to `status.hostIP`.

### Why is this needed?

NONE

sunya-ch:
#### What type of PR is this?

/kind feature

Fixes #284

```release-note

```

```docs

```

Signed-off-by: Sunyanan Choochotkaew <sunyanan.choochotkaew1@ibm.com>


pkg/kwok/controllers/node_controller.go
```diff
@@ -492,3 +492,12 @@ func (c *NodeController) Has(nodeName string) bool {
 func (c *NodeController) Size() int {
 	return c.nodesSets.Size()
 }
+
+// Get returns Has bool and corev1.Node if the node is existed
+func (c *NodeController) Get(nodeName string) (bool, corev1.Node) {
```
wzshiming, line 497:
```suggestion
func (c *NodeController) Get(nodeName string) (corev1.Node, bool) {
```

Usually, the bool that determines whether it exists is placed last

pkg/kwok/controllers/pod_controller.go
```diff
@@ -475,6 +478,19 @@ func (c *PodController) configurePod(pod *corev1.Pod, template string) ([]byte,
 		}
 		pod.Status.PodIP = ips[0]
 	}
+	has, node := c.nodeGetFunc(pod.Spec.NodeName)
```
wzshiming, line 481: Update the NodeIP function in the template

https://github.com/kubernetes-sigs/kwok/blob/8e2871ed268e8efbd3934eeb5432a9afe6844ac2/pkg/kwok/controllers/pod_controller.go#L126-L128

``` golang
"NodeIP": func(nodeName string) string {
	node, has := c.nodeGetFunc(nodeName)
	if has && len(node.HostIPs) > 0 {
		return node.HostIPs[0]
	}
	return n.nodeIP
},
```

and modify https://github.com/kubernetes-sigs/kwok/blob/8e2871ed268e8efbd3934eeb5432a9afe6844ac2/stages/pod-fast.yaml#L60-L61

``` template
      hostIP: {{ with .hostIP -}} 
        {{- . -}}
      {{- else -}}
        {{- NodeIP .metadata.name -}}
      {{- end }}

      podIP: {{ with .podIP -}}
        {{- . -}}
      {{- else -}}
        {{- if .spec.hostNetwork -}}
          {{- NodeIP .metadata.name -}}
        {{- else -}}
          {{- PodIP -}}
        {{- end -}}
      {{- end }}
```

This will be more flexible and versatile.

wzshiming (changes requested):

wzshiming:
The commit message may contain some special characters that cause `do-not-merge/invalid-commit-message`
> Indicates that a PR should not merge because it has an invalid commit message.

wzshiming:
https://github.com/kubernetes/community/blob/49fde191a905ae4bf165ef5850821fbe8fb36d96/contributors/guide/pull-requests.md?plain=1#L437

It looks like the title is causing the labels

/retitle Configure hostIP of Pod's status and support hostNetwork

pkg/kwok/controllers/pod_controller.go
sunya-ch: Thank you. Added in the amend change.

pkg/kwok/controllers/node_controller.go
```diff
@@ -483,6 +488,18 @@ func (c *NodeController) computePatchNode(node *corev1.Node, tpl string) ([]byte
 	})
 }
 
+// putNodeInfo puts node info (HostIPs)
+func (c *NodeController) putNodeInfo(node *corev1.Node) {
```
wzshiming, line 492: https://github.com/kubernetes/kubernetes/blob/1d02d014e8c1f0de84b0b58b2165548182815320/pkg/util/node/node.go#L67-L104

We should refer to the logic of how Kubelet handles addresses.

pkg/kwok/controllers/node_controller.go
sunya-ch: Thank you. Please check this change.
https://github.com/kubernetes-sigs/kwok/compare/657f81e2535805cb4df0a6955505509e46ab8403..c7817a162e7eee66c48009316f12fb8b40e6dde1

pkg/kwok/controllers/utils.go
```diff
@@ -162,12 +162,12 @@ func (p *parallelTasks) Wait() {
```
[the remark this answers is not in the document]
sunya-ch: Thank you. Please check this change: https://github.com/kubernetes-sigs/kwok/compare/67efb217e92e03e55be63d9a7a0825a2abcae712..657f81e2535805cb4df0a6955505509e46ab8403.

pkg/kwok/controllers/pod_controller.go
```diff
@@ -475,6 +482,19 @@ func (c *PodController) configurePod(pod *corev1.Pod, template string) ([]byte,
...
+			hostIP := nodeInfo.HostIPs[0]
+			if hostIP != "" {
+				pod.Status.HostIP = hostIP
+
+				if pod.Spec.HostNetwork {
+					pod.Status.PodIP = hostIP
+				}
+			}
+		}
+	}
```
wzshiming, line 497:
```suggestion
```

wzshiming (commented):
https://github.com/kubernetes-sigs/kwok/blob/cef98dea6bd74b1fcd5262f601bc293196c72fa5/test/kwok/kwok.test.sh#L111

It looks like it's almost ready to go.

Can we add a test here in case this part of the logic is accidentally broken

- For Pods without HostNetwork, PodIP and HostIP are not equal
- For Pods with HostNetwork, PodIP and HostIP are equal.
- Pod's HostIP is equal to the IP on its Node

FYI

- Get the Pod's hostIP `jq '.status.hostIP'`
- Get the Host's IP `jq '.status.addresses.[] | select( .type == "InternalIP" ) | .address'`

pkg/kwok/controllers/pod_controller.go
sunya-ch: I changed from `.metadata.name` to `.spec.nodeName`.

pkg/kwok/controllers/node_controller.go
sunya-ch: Thank you. I added the re-sort and also refer to v4 before v6.

pkg/kwok/controllers/pod_controller.go
sunya-ch: Thank. Removed.

pkg/kwok/controllers/node_controller.go
sunya-ch: Got timely test failure. Do you think is it related to re-sort logic when putting the nodeInfo? Should I optimize that? I think the kubelet logic needs only first ipv4 and first ipv6 (if exist) sorting from internal.

sunya-ch: According to `'.status.addresses.[] | select( .type == "InternalIP" ) | .address'`, 
should we optimize the logic to get only first ipv4/ipv6 which is InternalIP? Or should we keep all re-sorting logic?

wzshiming: Both should be feasible now, and my personal preference is to keep the sorting and the two IPs.

stages/pod-general.yaml
```diff
@@ -94,8 +95,22 @@ spec:
...
+      podIP: {{ with .podIP -}}
+        {{- . -}}
+      {{- else -}}
+        {{- if .spec.hostNetwork -}}
+          {{- NodeIPMap $nodeName -}}
+        {{- else -}}
+          {{- PodIP -}}
+        {{- end -}}
+      {{- end }}
+
```
wzshiming, line 113:
```suggestion
      {{ with .status }}
      hostIP: {{ with .hostIP -}} 
        {{- . -}}
      {{- else -}}
        {{- NodeIPMap $nodeName -}}
      {{- end }}

      podIP: {{ with .podIP -}}
        {{- . -}}
      {{- else -}}
        {{- if .spec.hostNetwork -}}
          {{- NodeIPMap $nodeName -}}
        {{- else -}}
          {{- PodIP -}}
        {{- end -}}
      {{- end }}
      {{- end }}
```

pkg/kwok/controllers/pod_controller.go
```diff
@@ -116,14 +118,22 @@ func NewPodController(conf PodControllerConfig) (*PodController, error) {
...
 		nodeHasFunc:                           conf.NodeHasFunc,
+		nodeGetFunc:                           conf.NodeGetFunc,
 		cronjob:                               cron.NewCron(),
 		lifecycle:                             lifecycles,
 		parallelTasks:                         newParallelTasks(conf.LockPodParallelism),
 		lockPodChan:                           make(chan *corev1.Pod),
 		recorder:                              conf.Recorder,
 	}
 	funcMap = template.FuncMap{
-		"NodeIP": func() string {
```
wzshiming, line 126: We should keep NodeIP for compatibility

wzshiming:
https://github.com/kubernetes-sigs/kwok/blob/cef98dea6bd74b1fcd5262f601bc293196c72fa5/pkg/kwok/controllers/pod_controller.go#L419

Additional, we should ignore HostNetwork in IP recycling

``` golang
if c.nodeHasFunc(pod.Spec.NodeName) && pod.Status.PodIP != pod.Status.HostIP && !pod.Spec.HostNetwork { 
```

pkg/kwok/controllers/pod_controller.go
sunya-ch: Fixed and added argument in `node-fast.yaml` too.

stages/pod-general.yaml
sunya-ch: Added `with .status` and changed NodeIPMap to NodeIP.

sunya-ch:
> https://github.com/kubernetes-sigs/kwok/blob/cef98dea6bd74b1fcd5262f601bc293196c72fa5/pkg/kwok/controllers/pod_controller.go#L419
> 
> Additional, we should ignore HostNetwork in IP recycling
> 
> ```go
> if c.nodeHasFunc(pod.Spec.NodeName) && pod.Status.PodIP != pod.Status.HostIP && !pod.Spec.HostNetwork { 
> ```

I think we may not add condition `pod.Status.PodIP != pod.Status.HostIP`. Adding this condition will affect the case `!pod.Spec.HostNetwork` and `c.enableCNI `. The enabled CNI might somehow set `pod.Status.PodIP=pod.Status.HostIP`. In that case, `cni.Remove` will be skipped.

stages/node-fast.yaml
```diff
@@ -48,7 +48,7 @@ spec:
       {{ with .status.addresses }}
       {{ YAML . 1 }}
       {{ else }}
-      {{ with NodeIP }}
+      {{ with NodeIP .metadata.name }}
```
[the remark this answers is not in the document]
sunya-ch: I have to make it the same because it's sharing the same FuncMap variable. 
Otherwise, I got the following error in the unit test.
```
{"time":"2023-02-20T19:48:15.204406+09:00","level":"ERROR","source":"/Users/aa404681/Documents/internal_ws/cni/kwok/pkg/kwok/controllers/pod_controller.go:281","msg":"Failed to configure pod","pod":"default/pod-with-hostNetwork","node":"node0","err":"template: _:47:5: executing \"_\" at \u003cNodeIP\u003e: wrong number of args for NodeIP: want 0 got 1"}
```
It's the reason I changed the key in pod controller to `NodeIPMap`.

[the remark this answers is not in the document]
sunya-ch: I have to make it the same because it's sharing the same FuncMap variable. 
Otherwise, I got the following error in the unit test.
```
{"time":"2023-02-20T19:48:15.204406+09:00","level":"ERROR","source":"_/cni/kwok/pkg/kwok/controllers/pod_controller.go:281","msg":"Failed to configure pod","pod":"default/pod-with-hostNetwork","node":"node0","err":"template: _:47:5: executing \"_\" at \u003cNodeIP\u003e: wrong number of args for NodeIP: want 0 got 1"}
```
It's the reason I changed the key in pod controller to `NodeIPMap`.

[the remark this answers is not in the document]
sunya-ch: I think it is the problem of unit test. funcMap are shared but the NewNodeController and NewPodController are called in a random order (unlike in `controller.go`) . The above error occurs because NewPodController is called before NewNodeController.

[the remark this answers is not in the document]
wzshiming: It seems to be a rather tangled place, do you have any good ideas?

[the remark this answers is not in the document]
sunya-ch: I think we should separate NodeIP and NodeIPMap because there is no need to check the NodeInfo map in the node controller. Different function should have different key. If we share the same key name, it will be overridden anyway. 

Another idea I have is,  instead of using a shared map, making a new function map when newRenderer for each controller.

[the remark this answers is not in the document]
sunya-ch: Another comment is about the unit test. Maybe we should also consider using test suite like ginkgo and define all controller before suite in the same way that we have in the controller.go.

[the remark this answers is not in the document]
wzshiming: > Another idea I have is, instead of using a shared map, making a new function map when newRenderer for each controller.

I am also more inclined to this approach

> Another comment is about the unit test. Maybe we should also consider using test suite like ginkgo and define all controller before suite in the same way that we have in the controller.go.

I agree with you, but this transition should not be done in this PR, we can start a new PR for this work.

`ginkgo` is usually used to write e2e or integration tests, currently e2e is written in shell, maybe we can use it to fill in the integration tests part.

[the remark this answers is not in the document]
sunya-ch: Made the changes by re-defining `commonFuncMap`.
https://github.com/kubernetes-sigs/kwok/compare/d484e263f4a8182cc3e79ae29117827a88d32aa0..17fe8647d6db8e631d96314d8359441948b74001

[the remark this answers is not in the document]
sunya-ch: I am not sure what is the purpose of shared funcMap (which is changed to commonFuncMap). 
Do you want it to be a global variable that always being added (common) or it is just a default variable for config value?  
For the former case, the `conf.FuncMap` should be controller-specific (i.e., not passing commonFuncMap) and copying commonFuncMap should be there. 
For the latter case, copying commonFuncMap (funcMap) is not needed. 

I will commit a new amend change regarding the purpose.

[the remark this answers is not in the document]
wzshiming: https://github.com/kubernetes-sigs/kwok/blob/8e2871ed268e8efbd3934eeb5432a9afe6844ac2/pkg/kwok/controllers/node_controller.go#L121

https://github.com/kubernetes-sigs/kwok/blob/8e2871ed268e8efbd3934eeb5432a9afe6844ac2/pkg/kwok/controllers/pod_controller.go#L125

This seems to be an omission from my previous Review where the global variable was changed, which is surprising, let's fix it.

```
funcMap := template.FuncMap{ 
```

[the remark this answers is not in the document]
sunya-ch: Made changes (I suppose you prefer the latter case. So, I renamed it to defaultFuncMap).
https://github.com/kubernetes-sigs/kwok/compare/17fe8647d6db8e631d96314d8359441948b74001..8943c6bcf4f05b5fba5f833ea0bb9905d1783a57

wzshiming:
Thank you, change LGTM


https://github.com/kubernetes-sigs/kwok/blob/cef98dea6bd74b1fcd5262f601bc293196c72fa5/test/kwok/kwok.test.sh#L111

Can we add a test to cover this change.
- For Pods without HostNetwork, PodIP and HostIP are not equal.
- For Pods with HostNetwork, PodIP and HostIP are equal.
- The Pod's HostIP is equal to the IP on its node

sunya-ch:
> Thank you, change LGTM
> 
> https://github.com/kubernetes-sigs/kwok/blob/cef98dea6bd74b1fcd5262f601bc293196c72fa5/test/kwok/kwok.test.sh#L111
> 
> Can we add a test to cover this change.
> 
> * For Pods without HostNetwork, PodIP and HostIP are not equal.
> * For Pods with HostNetwork, PodIP and HostIP are equal.
> * The Pod's HostIP is equal to the IP on its node

Amended  for changes.  However, I think we should not check the first condition because there may be the future case that supporting CNI like host-device that PodIP and Host IP are equal without HostNetwork. Each CNI should have its own test .

wzshiming:
> Amended for changes. However, I think we should not check the first condition because there may be the future case that supporting CNI like host-device that PodIP and Host IP are equal without HostNetwork. Each CNI should have its own test .

With the CNI case in a separate test https://github.com/kubernetes-sigs/kwok/tree/main/test/kwok-with-cni, this one is a standalone.

sunya-ch:
> > Amended for changes. However, I think we should not check the first condition because there may be the future case that supporting CNI like host-device that PodIP and Host IP are equal without HostNetwork. Each CNI should have its own test .
> 
> With the CNI case in a separate test https://github.com/kubernetes-sigs/kwok/tree/main/test/kwok-with-cni, this one is a standalone.

I think we should let that CNI test check Pod IP correctness (such as within CIDR) when hostNetwork=false (including For Pods without HostNetwork, PodIP and HostIP are not equal) and that could be in another PR.

wzshiming:
> I think we should let that CNI test check Pod IP correctness (such as within CIDR) when hostNetwork=false (including For Pods without HostNetwork, PodIP and HostIP are not equal) and that could be in another PR.

I agree with you, I also have such expectations

test/kwok/kwok.test.sh
```diff
@@ -65,6 +91,8 @@ function test_pod_running() {
     fi
   done
 
+  check_pod_status
```
wzshiming, line 94: Move to main

test_check_pod_status || failed+=("check_pod_status")

test/kwok/kwok.test.sh
```diff
@@ -54,6 +54,32 @@ function test_node_ready() {
   fi
 }
 
+# Check pod status
+function check_pod_status() {
+  for row in $(kubectl get po -ojson | jq -r '.items[] | @base64'); do
+      _jq() {
+      echo ${row} | base64 --decode | jq -r ${1}
+      }
+    podName=$(_jq '.metadata.name')
```
wzshiming, line 63:
```suggestion
    pod_name=$(_jq '.metadata.name')
```

At bash we generally use the snake case

wzshiming (changes requested):

test/kwok/kwok.test.sh
sunya-ch: Made changes [here](https://github.com/kubernetes-sigs/kwok/compare/42dd2e5d7ed0d857067ceb1187a1e5f39b37c4fa..6937431a58b0a2149d9fe50882102895d9b55002). Thanks.

test/kwok/kwok.test.sh
```diff
@@ -54,6 +54,32 @@ function test_node_ready() {
...
+  for row in $(kubectl get po -ojson | jq -r '.items[] | @base64'); do
+      _jq() {
+      echo ${row} | base64 --decode | jq -r ${1}
+      }
+    pod_name=$(_jq '.metadata.name')
+    node_name=$(_jq '.spec.nodeName')
+    host_network=$(_jq '.spec.hostNetwork')
+    host_ip=$(_jq '.status.hostIP')
+    pod_ip=$(_jq '.status.podIP')
+    node_ip=$(kubectl get nodes ${node_name} -ojson|jq '.status.addresses[]| select( .type == "InternalIP" )|.address'|tr -d '"')
```
wzshiming, line 68:
```suggestion
    node_ip=$(kubectl get nodes ${node_name} -ojson | jq -r '.status.addresses[] | 
 select( .type == "InternalIP" ) | .address')
```

test/kwok/kwok.test.sh
```diff
@@ -54,6 +54,32 @@ function test_node_ready() {
   fi
 }
 
+# Check pod status
+function test_check_pod_status() {
+  for row in $(kubectl get po -ojson | jq -r '.items[] | @base64'); do
+      _jq() {
+      echo ${row} | base64 --decode | jq -r ${1}
+      }
```
wzshiming, line 62:
```suggestion
  _jq() {
    echo ${row} | base64 --decode | jq -r ${1}
  }
  for row in $(kubectl get po -ojson | jq -r '.items[] | @base64'); do
```

test/kwok/kwok.test.sh
```diff
@@ -54,6 +54,32 @@ function test_node_ready() {
...
+    node_name=$(_jq '.spec.nodeName')
+    host_network=$(_jq '.spec.hostNetwork')
+    host_ip=$(_jq '.status.hostIP')
+    pod_ip=$(_jq '.status.podIP')
+    node_ip=$(kubectl get nodes ${node_name} -ojson|jq '.status.addresses[]| select( .type == "InternalIP" )|.address'|tr -d '"')
+    if [[ $host_ip != $node_ip ]]; then
+        echo "Error: ${pod_name} Pod: HostIP (${host_ip}) is not equal to the IP on its Node (${node_ip})"
+        return 1
+    fi
+    if [[ $hostNetwork == "true" ]]; then
```
wzshiming, line 73:
```suggestion
    if [[ $host_network == "true" ]]; then
```

wzshiming (changes requested):

test/kwok/kwok.test.sh
sunya-ch: Nice catch. This is critical.

sunya-ch: Applied the changes [here](https://github.com/kubernetes-sigs/kwok/compare/6937431a58b0a2149d9fe50882102895d9b55002..4849934ad12e13eb52cffd10cc563bf79e04f519).

test/kwok/kwok.test.sh
sunya-ch: Applied the changes [here](https://github.com/kubernetes-sigs/kwok/compare/6937431a58b0a2149d9fe50882102895d9b55002..4849934ad12e13eb52cffd10cc563bf79e04f519).

test/kwok/kwok.test.sh
sunya-ch: Applied the changes [here](https://github.com/kubernetes-sigs/kwok/compare/6937431a58b0a2149d9fe50882102895d9b55002..4849934ad12e13eb52cffd10cc563bf79e04f519).

wzshiming:
/approve
/lgtm

Thank you for your contribution

wzshiming (approved):

PR #305, opened 2023-02-20
check src exists and use abspath

helen-frank:
#### What type of PR is this?
/kind feature

- check whether the local binary exists.
- symlink cannot point to a file with a relative path

Fixes #

When debugging by specifying a local kwok-controller-binary, it is found that the cluster cannot be created normally. During troubleshooting, it is found that the files with relative paths will be linked by symbols when kwok-controller-binary use local files.

use `./kwokctl create cluster --name m1 --kwok-controller-binary ./kwok`

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/utils/file/download.go
+++ b/pkg/utils/file/download.go
@@ -114,7 +114,14 @@ func getCachePath(cacheDir, src string) (string, error) {
 	case "http", "https":
 		return path.Join(cacheDir, u.Scheme, u.Host, u.Path), nil
 	default:
-		return src, nil
+		src, err = path.Expand(src)
+		if err != nil {
+			return "", err
+		}
+		if _, err := os.Stat(src); err != nil {
+			return "", err
+		}
+		return src, err
 	}
 }
 
--- a/pkg/utils/path/path.go
+++ b/pkg/utils/path/path.go
@@ -17,25 +17,28 @@ limitations under the License.
 package path
 
 import (
+	"fmt"
 	"os"
+	"path/filepath"
 )
 
-// ExpandHome expands home directory in file paths.
-func ExpandHome(path string) string {
-	if len(path) == 0 {
-		return path
-	}
-	if path[0] != '~' {
-		return path
-	}
-	if len(path) > 1 && path[1] != '/' && path[1] != '\\' {
-		return path
+// Expand expands absolute directory in file paths.
+func Expand(path string) (string, error) {
+	if path == "" {
+		return "", fmt.Errorf("empty path")
 	}
 
-	home, err := os.UserHomeDir()
-	if err != nil {
-		return path
+	if path[0] == '~' {
+		home, err := os.UserHomeDir()
+		if err != nil {
+			return "", err
+		}
+		if len(path) == 1 {
+			path = home
+		} else if path[1] == '/' || path[1] == '\\' {
+			path = Join(home, path[2:])
+		}
 	}
 
-	return Join(home, path[1:])
+	return filepath.Abs(path)
 }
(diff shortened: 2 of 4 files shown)


pkg/utils/file/download.go
```diff
@@ -114,6 +114,16 @@ func getCachePath(cacheDir, src string) (string, error) {
 	case "http", "https":
 		return path.Join(cacheDir, u.Scheme, u.Host, u.Path), nil
 	default:
+		if _, err := os.Stat(src); err != nil {
```
wzshiming, line 117: https://github.com/kubernetes-sigs/kwok/blob/4cc651b6334ea0d3e8c484e0a47ca6186bc5686e/pkg/utils/path/path.go#L23-L41

We can add the function `Expand`, which expands the home and local paths, and then replaces function `ExpandHome`, use it here.

helen-frank: Nice idea, can I add the check file to it?

helen-frank: According to the function definition of this piece, it seems that there is no need to check whether the file exists.

pkg/utils/path/path.go
```diff
@@ -39,3 +40,15 @@ func ExpandHome(path string) string {
 
 	return Join(home, path[1:])
 }
+
+// Expand expands absolute directory in file paths.
+func Expand(path string) (string, error) {
```
wzshiming, line 45: We can merge the function of ExpandHome into Expand, both of them are actually the same function which is to expand the path.

pkg/utils/file/download.go
wzshiming: > According to the function definition of this piece, it seems that there is no need to check whether the file exists.

Yep

pkg/utils/path/path.go
```diff
@@ -17,25 +17,33 @@ limitations under the License.
...
-	home, err := os.UserHomeDir()
+	absPath, err := filepath.Abs(path)
 	if err != nil {
-		return path
+		return "", err
 	}
 
-	return Join(home, path[1:])
+	return absPath, nil
 }
```
wzshiming, line 49:
```suggestion
// Expand expands absolute directory in file paths.
func Expand(path string) (string, error) {
	if path == "" {
		return "", fmt.Errorf("empty path")
	}

	if path[0] == '~' {
		home, err := os.UserHomeDir()
		if err != nil {
			return "", err
		}

		if len(path) == 1 {
			path = home
		} else if path[1] == '/' || path[1] == '\\' {
			path = Join(home, path[2:])
		}
	}

	if !filepath.IsAbs(path) {
		absPath, err := filepath.Abs(path)
		if err != nil {
			return "", err
		}
		path = absPath
	}

	return path, nil
}
```

wzshiming: Can we add a test cover the Expand

helen-frank: ok, a little time

pkg/utils/file/download.go
```diff
@@ -114,7 +114,10 @@ func getCachePath(cacheDir, src string) (string, error) {
 	case "http", "https":
 		return path.Join(cacheDir, u.Scheme, u.Host, u.Path), nil
 	default:
-		return src, nil
+		if _, err := os.Stat(src); err != nil {
+			return "", err
+		}
+		return path.Expand(src)
```
wzshiming, line 120:
```suggestion
		src = path.Expand(src)
		if _, err := os.Stat(src); err != nil {
			return "", err
		}
		return src
```

wzshiming (changes requested):

pkg/utils/file/download.go
helen-frank: Safer approach, agreed

wzshiming:
/approve
/lgtm

Thank you for your contribution

PR #318, opened 2023-02-22
log: change not managed pod log to debug, print delay as string

zwpaper:

there are too many logs like `Skip pod, not manage` for pods running in nodes not managed by kwok, but users should not care about the pods not managed by kwok, so we should put it to Debug level.

the deplay time in logs is int, it is not possible to read by humans, changed to String, make it output like `1h2m3s`

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kwok/controllers/node_controller.go
+++ b/pkg/kwok/controllers/node_controller.go
@@ -359,7 +359,7 @@ func (c *NodeController) LockNode(ctx context.Context, node *corev1.Node) error 
 		return fmt.Errorf("stage match: %w", err)
 	}
 	if stage == nil {
-		logger.Info("Skip node",
+		logger.Debug("Skip node",
 			"reason", "not match any stages",
 		)
 		return nil
@@ -396,7 +396,7 @@ func (c *NodeController) LockNode(ctx context.Context, node *corev1.Node) error 
 				return
 			}
 			if patch == nil {
-				logger.Info("Skip node",
+				logger.Debug("Skip node",
 					"reason", "do not need to modify",
 				)
 			} else {
@@ -413,7 +413,7 @@ func (c *NodeController) LockNode(ctx context.Context, node *corev1.Node) error 
 		return nil
 	}
 
-	logger.Info("Delayed play stage",
+	logger.Debug("Delayed play stage",
 		"delay", delay,
 		"stage", stageName,
 	)
--- a/pkg/kwok/controllers/pod_controller.go
+++ b/pkg/kwok/controllers/pod_controller.go
@@ -245,7 +245,7 @@ func (c *PodController) LockPod(ctx context.Context, pod *corev1.Pod) error {
 		return fmt.Errorf("stage match: %w", err)
 	}
 	if stage == nil {
-		logger.Info("Skip pod",
+		logger.Debug("Skip pod",
 			"reason", "not match any stages",
 		)
 		return nil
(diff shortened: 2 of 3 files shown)


pkg/kwok/controllers/pod_controller.go
```diff
@@ -300,7 +300,7 @@ func (c *PodController) LockPod(ctx context.Context, pod *corev1.Pod) error {
 	}
 
 	logger.Info("Delayed play stage",
-		"delay", delay,
+		"delay", delay.String(),
```
wzshiming, line 303:
```suggestion
		"delay", delay,
		"delayString", delay.String(),
```

pkg/kwok/controllers/node_controller.go
```diff
@@ -414,7 +414,7 @@ func (c *NodeController) LockNode(ctx context.Context, node *corev1.Node) error
 	}
 
 	logger.Info("Delayed play stage",
-		"delay", delay,
+		"delay", delay.String(),
```
wzshiming, line 417:
```suggestion
		"delay", delay,
		"delayString", delay.String(),
```

pkg/kwok/controllers/pod_controller.go
```diff
@@ -416,8 +416,8 @@ func (c *PodController) WatchPods(ctx context.Context, lockChan chan<- *corev1.P
 					if c.needLockPod(pod) {
 						lockChan <- pod.DeepCopy()
 					} else {
-						logger.Info("Skip pod",
-							"reason", "not manage",
+						logger.Debug("Skip pod",
```
wzshiming, line 419: Maybe we should change all the Skip/Delayed messages to debug instead of just this one

zwpaper: previously I thought so too, but only changed this one when I saw others are logging under the managed nodes.

It is reasonable to change those skip logs, let me change them tonight

pkg/kwok/controllers/node_controller.go
```diff
@@ -413,8 +413,9 @@ func (c *NodeController) LockNode(ctx context.Context, node *corev1.Node) error
 		return nil
 	}
 
-	logger.Info("Delayed play stage",
+	logger.Debug("Delayed play stage",
 		"delay", delay,
+		"delayString", delay.String(),
```
wzshiming, line 418: https://github.com/kubernetes-sigs/kwok/blob/122f36ce87128371d2b9bbcff43837d764b35e37/pkg/log/logger.go#L58-L68

``` golang
			if a.Value.Kind() == slog.DurationKind {
				if t, ok := a.Value.Any().(time.Duration); ok {
					return slog.Any(a.Key, DurationFormat{
						Nanosecond: int64(t),
						Human:      t.String(),
					})
				}
			}
```

I'm thinking deeply about this issue, and adding in this is more appropriate.

pkg/log/logger.go
```diff
@@ -29,6 +30,12 @@ import (
 // IsTerminal returns true if the given file descriptor is a terminal.
 var IsTerminal = term.IsTerminal
 
+// DurationFormat is the format used to print time.Duration in both nanosecond and string.
+type DurationFormat struct {
+	Nanosecond int64
+	Human      string
```
wzshiming, line 36:
```suggestion
	Nanosecond int64  `json:"nanosecond"`
	Human      string `json:"human"`
```

PR #324, opened 2023-02-23
Set/remove current-context after cluster create/deletion

### LINKED ISSUE #320: [kwokctl] set/remove `current-context` after cluster create/deletion

When a user creates a cluster using `kwokctl`, it's better to automatically switch the `current-context` to the newly created one.

```
⇒  k ctx -c
kwok-k2

⇒  kwokctl create cluster --name k3
...
Cluster is created                                                                                                                               cluster=k3 elapsed=1s
You can now use your cluster with:

    kubectl config use-context kwok-k3

Thanks for using kwok!

⇒  k ctx -c
kwok-k2 <- I'd expect it switches to kwok-k3 automatically
```

The same story applies to cluster deletion. I'd like to unset the current-context.

### Why is this needed?

For better UX. When a user creates/deletes a cluster, their instinct is to use/not use that cluster right away.

qihuang0:
#### What type of PR is this?
/kind feature

With this commit， we can set/remove current-context after cluster create/deletion

Fixes #320 

```release-note

```

```docs

```

### DIFF
--- a/kwok.yaml
+++ b/kwok.yaml
@@ -1,0 +1,7 @@
+kind: KwokConfiguration
+apiVersion: kwok.x-k8s.io/v1alpha1
+options:
+---
+kind: KwokctlConfiguration
+apiVersion: kwok.x-k8s.io/v1alpha1
+options:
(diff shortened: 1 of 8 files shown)


wzshiming:
Hi, do you have time to continue this work, I intend to take it forward.

qihuang0:
> Hi, do you have time to continue this work, I intend to take it forward.

I'm sorry, but I've been really busy lately and have decided to give up this work

Issue #261, opened 2023-02-07
[kwokctl] Support port exposure to components


wzshiming:
### What would you like to be added?

- [x] `--controller-port` #263
- [x] `--etcd-port` #325
- [ ] `--kube-controller-manager-port` #383
- [x] `--kube-scheduler-port` #369

like `--prometheus-port`

### Why is this needed?

NONE

wzshiming:
@nikola-jokic
Amazing, Thank you for completing all the tasks in this issue 🎉🎉🎉

PR #325, opened 2023-02-23
[kwokctl] Support port exposure for etcd-port

nikola-jokic:
#### What type of PR is this?

/kind feature

Allow access to the etcd from the host machine

Related to #261 (etcd-port)

This is currently using nc to test connectivity. The `curl` command to query a version will fail on kind since the host does not have a valid cert. This might cause failure on binary runtime CI because the nc is not installed.

### DIFF
--- a/pkg/kwokctl/components/etcd.go
+++ b/pkg/kwokctl/components/etcd.go
@@ -36,32 +36,30 @@ type BuildEtcdComponentConfig struct {
 
 // BuildEtcdComponent builds an etcd component.
 func BuildEtcdComponent(conf BuildEtcdComponentConfig) (component internalversion.Component, err error) {
+	exposePeerPort := true
 	if conf.PeerPort == 0 {
 		conf.PeerPort = 2380
+		exposePeerPort = false
 	}
+	exposePort := true
 	if conf.Port == 0 {
 		conf.Port = 2379
+		exposePort = false
 	}
 	if conf.Address == "" {
 		conf.Address = publicAddress
 	}
-	etcdPeerPortStr := format.String(conf.PeerPort)
-	etcdClientPortStr := format.String(conf.Port)
+
+	var volumes []internalversion.Volume
+	var ports []internalversion.Port
 
 	etcdArgs := []string{
 		"--name=node0",
-		"--initial-advertise-peer-urls=http://" + conf.Address + ":" + etcdPeerPortStr,
-		"--listen-peer-urls=http://" + conf.Address + ":" + etcdPeerPortStr,
-		"--advertise-client-urls=http://" + conf.Address + ":" + etcdClientPortStr,
-		"--listen-client-urls=http://" + conf.Address + ":" + etcdClientPortStr,
-		"--initial-cluster=node0=http://" + conf.Address + ":" + etcdPeerPortStr,
 		"--auto-compaction-retention=1",
 		"--quota-backend-bytes=8589934592",
 	}
 
 	inContainer := conf.Image != ""
-	var volumes []internalversion.Volume
-
 	if inContainer {
 		// TODO: use a volume for the data path
 		// volumes = append(volumes,
@@ -73,9 +71,42 @@ func BuildEtcdComponent(conf BuildEtcdComponentConfig) (component internalversio
 		etcdArgs = append(etcdArgs,
 			"--data-dir=/etcd-data",
 		)
+
+		if exposePeerPort {
+			ports = append(
+				ports,
+				internalversion.Port{
+					HostPort: conf.PeerPort,
+					Port:     2380,
+				},
+			)
+		}
+		if exposePort {
+			ports = append(
+				ports,
+				internalversion.Port{
+					HostPort: conf.Port,
+					Port:     2379,
+				},
+			)
+		}
+		etcdArgs = append(etcdArgs,
+			"--initial-advertise-peer-urls=http://"+conf.Address+":2380",
+			"--listen-peer-urls=http://"+conf.Address+":2380",
+			"--advertise-client-urls=http://"+conf.Address+":2379",
+			"--listen-client-urls=http://"+conf.Address+":2379",
+			"--initial-cluster=node0=http://"+conf.Address+":2380",
+		)
 	} else {
+		etcdPeerPortStr := format.String(conf.PeerPort)
+		etcdClientPortStr := format.String(conf.Port)
 		etcdArgs = append(etcdArgs,
 			"--data-dir="+conf.DataPath,
+			"--initial-advertise-peer-urls=http://"+conf.Address+":"+etcdPeerPortStr,
+			"--listen-peer-urls=http://"+conf.Address+":"+etcdPeerPortStr,
+			"--advertise-client-urls=http://"+conf.Address+":"+etcdClientPortStr,
+			"--listen-client-urls=http://"+conf.Address+":"+etcdClientPortStr,
+			"--initial-cluster=node0=http://"+conf.Address+":"+etcdPeerPortStr,
 		)
 	}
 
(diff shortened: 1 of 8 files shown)


wzshiming:
/retitle [kwokctl] Support port exposure for etcd-port

pkg/kwokctl/runtime/kind/kind.yaml.tpl
```diff
@@ -23,6 +23,12 @@ nodes:
     listenAddress: "0.0.0.0"
     protocol: TCP
   {{ end }}
+  {{ if .EtcdPort }}
+  - containerPort: {{ .EtcdPort }}
```
wzshiming, line 27:
```suggestion
  - containerPort: 2379
```

The port of Etcd in Kind configuration is here, 

https://github.com/kubernetes-sigs/kwok/blob/cbe4878f0a7a41b77fca5c9840be976747c0fefc/pkg/kwokctl/runtime/kind/prometheus_deployment.yaml.tpl#L70

pkg/kwokctl/runtime/compose/prometheus.yaml.tpl
```diff
@@ -29,7 +29,7 @@ scrape_configs:
   enable_http2: true
   static_configs:
   - targets:
-    - "{{ .ProjectName }}-etcd:2379"
+    - "{{ .ProjectName }}-etcd:{{or .EtcdPort "2379" }}"
```
wzshiming, line 32:
```suggestion
    - "{{ .ProjectName }}-etcd:{{ or .EtcdPort 2379 }}"
```

test/kwokctl/kwokctl_workable_test.sh
```diff
@@ -132,6 +132,19 @@ function test_kwok_controller_port() {
     fi
 }
 
+function test_etcd_port() {
+    local result
+    nc -w 1 -z 127.0.0.1 2400
```
wzshiming, line 137:
```
$ curl http://127.0.0.1:2400/health
{"health":"true","reason":""}
```

I remember this port is for http, maybe we curl it directly...

nikola-jokic: The problem is when the runtime is kind. Kind will start the etcd cluster with the certificate, causing http or https request to fail on a handshake... 
Not sure how to work around that issue without making it overly complicated

test/kwokctl/kwokctl_workable_test.sh
```diff
@@ -132,6 +132,18 @@ function test_kwok_controller_port() {
     fi
 }
 
+function test_etcd_port() {
+    local result
+    result=$(curl -s http://127.0.0.1:2400/health)
+
+    if [[ "$(echo "${result}" | grep -o '"health"' | wc -c)" = 0 ]]; then
```
nikola-jokic, line 139: Locally, the `$result` would be `{"health":"true","reason":""}`, but on CI, the result was:
`{"health: "true"}` (also with space between `:` and `"`.
So I thought just being able to fetch `"health"` was good enough sign that we are able to talk to the etcd

wzshiming: I think it's OK

pkg/kwokctl/components/etcd.go
```diff
@@ -73,6 +77,13 @@ func BuildEtcdComponent(conf BuildEtcdComponentConfig) (component internalversio
 		etcdArgs = append(etcdArgs,
 			"--data-dir=/etcd-data",
 		)
+
+		if exposePort {
+			ports = append(ports, internalversion.Port{
+				Port:     conf.Port,
```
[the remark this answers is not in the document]
nikola-jokic: I think that would be great! The difference between the etcd port and the rest of it is that it defaults to the value used by the api server. But if that is abstracted for docker runtime, for example, you don't need to care about what is the actual port inside the container. Brilliant!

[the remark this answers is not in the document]
wzshiming: Ohhh, I forgot about that, so we'll make the other components behave in line with this in the future. 😄

wzshiming:
There's another test that didn't pass that doesn't look like flake.

pkg/kwokctl/components/etcd.go
```diff
@@ -73,9 +67,35 @@ func BuildEtcdComponent(conf BuildEtcdComponentConfig) (component internalversio
 		etcdArgs = append(etcdArgs,
 			"--data-dir=/etcd-data",
 		)
+
+		ports = append(
+			ports,
+			internalversion.Port{
+				HostPort: conf.Port,
```
nikola-jokic, line 74: Not sure how to approach this one...
So if you start the cluster and not specify the port, we would expose this one, which is the default port 2379.
Should we default, or should we leave it to 0 in this case? This might cause problems on systems having etcd running as a service, if they do not specify the `--etcd-port`. I would lean more toward 0, but I can't really make a decision here :smile:

wzshiming: Add a condition to this exposed port

``` golang
	exposePeerPort := true
	if conf.PeerPort == 0 {
		conf.PeerPort = 2380
		exposePeerPort = false
	}
	exposePort := true
	if conf.Port == 0 {
		conf.Port = 2379
		exposePort = false
	}

```

``` golang
	if exposePort {
		ports = append(
			ports,
			internalversion.Port{
				HostPort: conf.Port,
				Port:     2379,
			},
		)
	}
	if exposePeerPort {
		ports = append(
			ports,
			internalversion.Port{
				HostPort: conf.PeerPort,
				Port:     2380,
			},
		)
	}
```

test/kwokctl/kwokctl_workable_test.sh
```diff
@@ -140,6 +152,9 @@ function main() {
     echo "Testing workable on ${KWOK_RUNTIME} for ${release}"
     name="cluster-${KWOK_RUNTIME}-${release//./-}"
     test_create_cluster "${release}" "${name}" || failed+=("create_cluster_${name}")
+    if ! [[ "${KWOK_RUNTIME}" = "kind" ]]; then
```
wzshiming, line 155:
```suggestion
    if [[ "${KWOK_RUNTIME}" != "kind" ]]; then
```

pkg/kwokctl/components/etcd.go
```diff
@@ -36,32 +36,30 @@ type BuildEtcdComponentConfig struct {
 
 // BuildEtcdComponent builds an etcd component.
 func BuildEtcdComponent(conf BuildEtcdComponentConfig) (component internalversion.Component, err error) {
+	exposePeerPort := true
```
wzshiming, line 39: nit: Maybe we should add a flag for this

wzshiming (commented):
LGTM, squash commits, then merge it.

pkg/kwokctl/components/etcd.go
nikola-jokic: I think that could be tricky for following reasons:
- On binary runtime, you would have to have port available if I am not mistaking
- On other runtimes, if you specify --etcd-port, it is going to be exposed, but under the hood, it is still going to use 2379, so basically, by specifying the flag you are exposing it... 
If flag is added to expose, then on docker and kind runtimes, your --etcd-port is useless if --expose-port or similar flag is not set
Please correct me if I am wrong :smile:

wzshiming: Sorry for the misunderstanding caused by my comment, I meant whether to add `--etcd-peer-port` to expose the peer port.

nikola-jokic: Oh, sure, I can do that :relaxed: on it

wzshiming: NVM the comment, peer port is used for etcd clusters, but should not be needed in our kwok usage scenario.

wzshiming:
This should be a known flake test, I'll try to fix it when I have time. 😭

PR #356, opened 2023-03-07
[Docs] Rewrite kwok-in-local.md

### LINKED ISSUE #333: [doc] Unclear what local refers to in installation docs

https://kwok.sigs.k8s.io/docs/user/kwok-in-local/ this doc doesn't make it clear what "in the local" means. For example, can it work on a Mac or Windows WSL where Docker is installed and running?  Or does it need a Kubernetes cluster?

e.g. If I run the command specified directly, it tries to talk to a local kind cluster (my current-context set on kubectl) but it fails because it's down

```
$ kwok --kubeconfig=~/.kube/config  ... # the command in the docs

  ERROR Failed to list nodes                                    err="Get \"https://127.0.0.1:45679/api/v1/nodes?limit=1\": dial tcp 127.0.0.1:45679: connect: connection refused"
  ERROR Failed to list nodes                                    err="Get \"https://127.0.0.1:45679/api/v1/nodes?limit=1\": dial tcp 127.0.0.1:45679: connect: connection refused"
  ...
```

### Why is this needed?

User experience.

This output makes me think this tool clearly needs a local Kubernetes cluster to be able to run, because it's trying to use the local Kubernetes cluster.

* But isn't the tool for bringing up a fake apiserver plus some components locally? Why does it need Kubernetes to do that (like `kind`)?
* And how is this different than ["Kwok in a Cluster"](https://kwok.sigs.k8s.io/docs/user/kwok-in-cluster/) if it needs a k8s cluster to run anyway?

Long story short, I think I should be using `kwokctl create cluster` command, but why is the "Kwok on local" page telling me to install it to my Mac via Homebrew and use `kwok` executable instead?

windsonsea:
#### What type of PR is this?

/kind documentation

Describe the local environment to run `kwok`

Fixes #333 

```release-note
NONE
```


ahmetb:
Sorry I'm not sure how this addresses #333, I still can't tell what "local" means and how's it different than "kwok in a Cluster". :)

Huang-Wei:
The term `kwok in Local` is misleading... It _primarily_ refers to running `kwok` in binary mode, which is a good companion of `hack/local-up-cluster.sh` on Mac in two aspects:

- Darwin kubelet is not available (yet), so `hack/local-up-cluster.sh` on Mac gives you a bare-bone k8s control plane
- In this k8s setup, there is no kubelet to accommodate a k8s deployment of kwok controller. So options are either running kwok as a binary or raw container.

@wzshiming Given this page cause more confusion than benefits. I'd suggest simply removing it. Or, in somewhere else, reword it to be a page like "running kwok, the hard way" where it focus on explaining running kwok in some unpopular way.

wzshiming:
The original intent of this page was to describe how kwok works as a Node and Pod controller, running locally through kubeconfig for the cluster.

I think a few points need to be added to the
- Description kwok is just a controller
- In general kwok provides an image, which can be easily deployed in a cluster by working in-cluster mode for the cluster
- Running in Local here means that kwok works outside the cluster via kubeconfig for the cluster


This title may be misleading, let me clarify it and think of a new one.

site/content/en/docs/user/kwok-in-local.md
```diff
@@ -6,8 +6,33 @@ This document walks you through how to run `kwok` in your local.
 
 {{< /hint >}}
 
-After you have [`kwok` installed]({{< relref "/docs/user/install" >}}) in a local computer,
-you can run `kwok` immediately for clusters and maintain their heartbeats:
+## Create clusters
```
wzshiming, line 9:
``` readme

# Running `kwok` out of cluster

{{< hint "info" >}}

This document walks you through how to run `kwok` out of cluster for a Kubernetes cluster.

{{< /hint >}}

## Prerequisites

Firstly, we need to have a Kubernetes cluster, and the `kwok` command-line tool must be available in your `PATH`.

## Running

Next, we can run `kwok` out of cluster by using the `--kubeconfig` flag.

```bash
kwok \
  --kubeconfig=~/.kube/config \
  --manage-all-nodes=false \
  --manage-nodes-with-annotation-selector=kwok.x-k8s.io/node=fake \
  --manage-nodes-with-label-selector= \
  --disregard-status-with-annotation-selector=kwok.x-k8s.io/status=custom \
  --disregard-status-with-label-selector= \
  --cidr=10.0.0.1/24 \
  --node-ip=10.0.0.1
\```

Finally, we can see the `kwok` is running out of cluster for the Kubernetes cluster.

## Next Steps

Now, we can use `kwok` to [manage nodes and pods] in the Kubernetes cluster.

[manage nodes and pods]: {{< relref "/docs/user/kwok-manage-nodes-and-pods" >}}
```

wzshiming: Change the path to `kwok-out-cluster` ?

site/content/en/docs/user/kwok-out-cluster.md
```diff
@@ -0,0 +1,36 @@
+
```
wzshiming, line 1: Make the original page redirect here

```suggestion
---
aliases:
  - /docs/user/kwok-in-local/
---

```

wzshiming (approved):
/approve
/lgtm

PR #369, opened 2023-03-08
[kwokctl] Support port exposure for kube-scheduler

nikola-jokic:
#### What type of PR is this?

/kind feature

Allow access to the kube-scheduler from the host machine

Related to #261  (kube-scheduler-port)

### DIFF
--- a/test/kwokctl/kwokctl_restart_test.sh
+++ b/test/kwokctl/kwokctl_restart_test.sh
@@ -54,7 +54,7 @@ function test_delete_cluster() {
 
 function test_prometheus() {
   local targets
-  for ((i = 0; i < 30; i++)); do
+  for ((i = 0; i < 60; i++)); do
     targets="$(curl -s http://127.0.0.1:9090/api/v1/targets)"
     if [[ "$(echo "${targets}" | grep -o '"health":"up"' | wc -l)" -ge 6 ]]; then
       break
--- a/test/kwokctl/kwokctl_workable_test.sh
+++ b/test/kwokctl/kwokctl_workable_test.sh
@@ -66,7 +66,7 @@ function test_create_cluster() {
   local targets
   local i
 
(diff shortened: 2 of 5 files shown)


nikola-jokic:
This change does not work on kind right now.
Kind uses in-cluster config, and not kube scheduler component.
Should we source the scheduler config and modify it to use NodePort? https://kubernetes.io/docs/reference/scheduling/config/#scheduling-plugins

wzshiming:
> This change does not work on kind right now. Kind uses in-cluster config, and not kube scheduler component. Should we source the scheduler config and modify it to use NodePort? https://kubernetes.io/docs/reference/scheduling/config/#scheduling-plugins

I agree with you, let's just explain this in the flag and documentation

I'm considering whether to allow more expected behavior differences in the Kind runtime in the future, it's a bit of a workload to maintain consistency with other runtime behaviors.

test/kwokctl/kwokctl_restart_test.sh
```diff
@@ -54,7 +54,7 @@ function test_delete_cluster() {
 
 function test_prometheus() {
   local targets
-  for ((i = 0; i < 30; i++)); do
+  for ((i = 0; i < 60; i++)); do
```
nikola-jokic, line 57: The scheduler was slow to get back up after the restart on nerdctl :relaxed:

test/kwokctl/kwokctl_workable_test.sh
```diff
@@ -144,6 +144,28 @@ function test_etcd_port() {
...
+    minor="${minor%.*}"
+
+    local proto="https"
+    if [[  $minor -le 12 ]]; then
+        proto="http"
+    fi
+
+    result=$(curl -s -k "${proto}://127.0.0.1:10250/healthz")
+
+    if [[ ! "${result}" = "ok" ]]; then
```
wzshiming, line 161:
```suggestion
    if [[ "${result}" != "ok" ]]; then
```

wzshiming (commented):
Others LGTM, and squash the commits.

wzshiming (approved):

PR #372, opened 2023-03-09
Clean up kwok init

sologgfun:
#### What type of PR is this?
/kind cleanup

The meaning of the parameter "Master" is unclear, and the descriptor uses the word "server".

```release-note
```

```docs

```

### DIFF
--- a/pkg/kwok/cmd/root.go
+++ b/pkg/kwok/cmd/root.go
@@ -84,7 +84,7 @@ func NewCommand(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVar(&flags.Options.DisregardStatusWithAnnotationSelector, "disregard-status-with-annotation-selector", flags.Options.DisregardStatusWithAnnotationSelector, "All node/pod status excluding the ones that match the annotation selector will be watched and managed.")
 	cmd.Flags().StringVar(&flags.Options.DisregardStatusWithLabelSelector, "disregard-status-with-label-selector", flags.Options.DisregardStatusWithLabelSelector, "All node/pod status excluding the ones that match the label selector will be watched and managed.")
 	cmd.Flags().StringVar(&flags.Kubeconfig, "kubeconfig", flags.Kubeconfig, "Path to the kubeconfig file to use")
-	cmd.Flags().StringVar(&flags.Master, "master", flags.Master, "Server is the address of the kubernetes cluster")
+	cmd.Flags().StringVar(&flags.Master, "master", flags.Master, "The address of the Kubernetes API server (overrides any value in kubeconfig).")
 	cmd.Flags().StringVar(&flags.Options.ServerAddress, "server-address", flags.Options.ServerAddress, "Address to expose health and metrics on")
 
 	cmd.Flags().BoolVar(&flags.Options.EnableCNI, "experimental-enable-cni", flags.Options.EnableCNI, "Experimental support for getting pod ip from CNI, for CNI-related components, Only works with Linux")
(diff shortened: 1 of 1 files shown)


pkg/kwok/cmd/root.go
```diff
@@ -84,7 +84,7 @@ func NewCommand(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringVar(&flags.Options.DisregardStatusWithAnnotationSelector, "disregard-status-with-annotation-selector", flags.Options.DisregardStatusWithAnnotationSelector, "All node/pod status excluding the ones that match the annotation selector will be watched and managed.")
 	cmd.Flags().StringVar(&flags.Options.DisregardStatusWithLabelSelector, "disregard-status-with-label-selector", flags.Options.DisregardStatusWithLabelSelector, "All node/pod status excluding the ones that match the label selector will be watched and managed.")
 	cmd.Flags().StringVar(&flags.Kubeconfig, "kubeconfig", flags.Kubeconfig, "Path to the kubeconfig file to use")
-	cmd.Flags().StringVar(&flags.Master, "master", flags.Master, "Server is the address of the kubernetes cluster")
+	cmd.Flags().StringVar(&flags.MasterUrl, "master-url", flags.MasterUrl, "MasterUrl is the address of the kubernetes cluster (https://hostname:port).")
```
wzshiming, line 87: This flag is the one that comes with the kube cli and should not be changed

wzshiming (changes requested):

sologgfun:
In addition, I have a small question.
```
// /pkg/kwon/cmd/root.go
RunE: func(cmd *cobra.Command, args []string) error {
	return runE(cmd.Context(), flags)
}

stagesData := config.FilterWithTypeFromContext[*internalversion.Stage](ctx)
```

The above code retrieves stagesData from ctx, but ctx is a new context, not the return parameter of `ctx, err := config.InitFlags(ctx, flagset)`
So where set the data to ctx？

pkg/kwok/cmd/root.go
sologgfun: ok,but the description `Server is the address of the kubernetes cluster` looks like a parameter named `Server`.
And I find the same description field in k8s named `Server`

wzshiming: Sync to `kube-controller-manager`

https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/

```
--master string
The address of the Kubernetes API server (overrides any value in kubeconfig).
```

wzshiming (changes requested):

pkg/kwok/cmd/root.go
sologgfun: Okay, I have edited and pushed the new code.

pkg/kwok/cmd/root.go
```diff
@@ -267,3 +249,29 @@ func filterStages(stages []*internalversion.Stage, apiGroup, kind string) []*int
 		return stage.Spec.ResourceRef.APIGroup == apiGroup && stage.Spec.ResourceRef.Kind == kind
 	})
 }
+
+func waitForReady(ctx context.Context, clientset kubernetes.Interface, logger *log.Logger) error {
```
wzshiming, line 253:
```suggestion
func waitForReady(ctx context.Context, clientset kubernetes.Interface) error {
	logger := log.FromContext(ctx)
```

sologgfun: ok, It`s looks better~

sologgfun:
I have rewritten the redundant code. Please take another look. Thank you.

wzshiming (approved):

PR #380, opened 2023-03-15
Process not using parent context

lixingchenDaoCloud:
#### What type of PR is this?

/kind feature

The process is not using the parent context, process will be terminated when the context is canceled.

Fixes #

NONE


pkg/utils/exec/cmd.go
```diff
@@ -72,8 +72,7 @@ func ForkExec(ctx context.Context, dir string, name string, arg ...string) error
 		return fmt.Errorf("write cmdline file %s: %w", cmdlinePath, err)
 	}
 
-	subCtx := context.Background()
```
wzshiming, line 75: This context is intentionally new, to avoid the cancellation of forked processes on binary runtime

PR #381, opened 2023-03-16
[kwokctl] Support k8s format for snapshot

### LINKED ISSUE #32: [kwokctl] Support k8s format for snapshot
```
kwokctl snapshot save --path snapshot.yaml --format=k8s
kwokctl snapshot restore --path snapshot.yaml --format=k8s
```

`Save --format=k8s`: means to convert k8s yaml to protobuf format and put it in the etcd database.
`Restore --format=k8s`: means to get the data in protobuf format from etcd and convert it to yaml.

Non goal: Ability to filtering of namespace, resource, label, etc.

wzshiming:
#### What type of PR is this?

/kind feature

Fixes #32

```release-note

```

```docs

```

### DIFF
--- a/pkg/kwokctl/snapshot/save.go
+++ b/pkg/kwokctl/snapshot/save.go
@@ -1,0 +1,35 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package snapshot
+
+import (
+	"context"
+	"io"
+	"strings"
+
+	"sigs.k8s.io/kwok/pkg/utils/exec"
+)
+
+// Save saves the snapshot of cluster
+func Save(ctx context.Context, rt Runtime, w io.Writer, filters []string) error {
+	err := rt.KubectlInCluster(exec.WithWriteTo(ctx, w), "get", strings.Join(filters, ","), "-o=yaml", "--all-namespaces")
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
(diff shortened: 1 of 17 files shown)


pkg/kwokctl/snapshot/save.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+	"io"
+	"strings"
+
+	"sigs.k8s.io/kwok/pkg/utils/exec"
+)
+
+// Save saves the snapshot of cluster
+func Save(ctx context.Context, rt Runtime, w io.Writer, filters []string) error {
+	err := rt.KubectlInCluster(exec.WithWriteTo(ctx, w), "get", strings.Join(filters, ","), "-o=yaml", "--all-namespaces")
```
Huang-Wei, line 29: in practice, it's likely to fail due to hitting QPS limits, right?

wzshiming: I've tried the latest kubectl, which supports chunking without failing even with very large amounts of data. 

However, the QPS limit causes it to be really slow, so I will try to remove the client/server QPS limit in the future to make this work better.

```
--chunk-size 500  Return large lists in chunks rather than all at once. Pass 0 to disable. This flag is beta and may change in the future.
```

Huang-Wei: Let's put it to this week's meeting discussion. My understanding is that `--chunk-size` is not the same thing as "client-go's paged query utility" (where it can define pageSize and retry logic programmatically), so IMO the latter impl. is what we need to pursue.

wzshiming: That's great, I have documented the relevant code links in advance as additional information for the discussion.

https://github.com/kubernetes/kubernetes/blob/d89d5ab2680bc74fe4487ad71e514f4e0812d9ce/staging/src/k8s.io/kubectl/pkg/cmd/get/get.go#L465

https://github.com/kubernetes/kubernetes/blob/d89d5ab2680bc74fe4487ad71e514f4e0812d9ce/staging/src/k8s.io/cli-runtime/pkg/resource/builder.go#L555

https://github.com/kubernetes/kubernetes/blob/c56be1fa9f1a0329c08cd0e2f9fc46dd903fc9ca/staging/src/k8s.io/cli-runtime/pkg/resource/selector.go#L54

Huang-Wei: Yes, that is correct. In my local fork, I utilized client-go's paged query to take a snapshot of the API objects. I had planned to contribute it back for really a while... hopefully I can find some time to make it soon.

PR #417, opened 2023-03-24
[kwok] Provided service for logs

### LINKED ISSUE #217: [kwok] Provided service for `logs`
``` yaml
# Match pods
kind: ClusterLogs
metadata:
  name: cluster-logs-rules
spec:
  selector:
    matchNamespaces:
      - podNamespace
    matchNames:
      - podName
  logs: []
--- 
# Just for a Pod
kind: Logs
metadata:
  name: podName
  namespace: podNamespace
spec:
  logs:
    - containers:
        - containerName

      # The first response is the contents of the logsFile
      # TODO: I'm not sure how to do support the --since-time, --since, --tail, --previous
      logsFile: /tmp/kwok.log

      # Follow up if true
      follow: true

```

/kind feature

nikola-jokic:
#### What type of PR is this?

/kind feature

Fixes #217


pkg/kwok/server/debugging_logs.go
```diff
@@ -89,3 +126,424 @@ func (s *Server) getContainerLogs(request *restful.Request, response *restful.Re
...
+	if apiOpts.SinceSeconds != nil {
+		opts.since = now.Add(-time.Duration(*apiOpts.SinceSeconds) * time.Second)
+	}
+	if apiOpts.SinceTime != nil && apiOpts.SinceTime.After(opts.since) {
+		opts.since = apiOpts.SinceTime.Time
+	}
+	return opts
+}
+
+func readLogs(ctx context.Context, logInfo *internalversion.Log, opts *logOptions, stdout, stderr io.Writer) error {
```
nikola-jokic, line 216: Most of this implementation comes from https://github.com/kubernetes/kubernetes/blob/0c62b122c02bff9131b6db960042150a3638d3f3/pkg/kubelet/kuberuntime/logs/logs.go#L283

I think it conforms nicely with the way the kubelet is working with logs, but I might be wrong here

test/kwokctl/kwokctl_logs_test.sh
```diff
@@ -0,0 +1,111 @@
...
+    echo "Error: Cluster ${name} creation failed"
+    exit 1
+  fi
+}
+
+function test_logs() {
+  local name="${1}"
+  local namespace="${2}"
+  local target="${3}"
+  # echo kwokctl --name "${name}" kubectl -n "${namespace}" logs "${target}" 2>&1 > /dev/null
```
nikola-jokic, line 53: Not sure how to test this. I considered adding unit tests, but I wanted to add this to start a discussion.

I read this section: https://kubernetes.io/docs/concepts/cluster-administration/logging/#how-nodes-handle-container-logs

My best guess is to create a log file that is either in CRI logging format or in docker logging format
That is feasible in binary runtime, but it would be tricky to add logs to the container

wzshiming: > Not sure how to test this. I considered adding unit tests, but I wanted to add this to start a discussion.

In my opinion, e2e can be done like this

```
echo "Multi line log" > input.log
kubectl logs --tail=1 -f podname > output.log &
echo "append a line log" >> input.log
# Then check that the last two lines of `output.log` are as expected
```

> My best guess is to create a log file that is either in CRI logging format or in docker logging format
That is feasible in binary runtime, but it would be tricky to add logs to the container

I think we need to actively mount the log files used by ClusterLogs/Logs into the container when we create the cluster using kwokctl?

pkg/kwok/server/debugging_logs.go
wzshiming: I think that's enough 😄

pkg/kwok/server/debugging_logs.go
```diff
@@ -89,3 +126,424 @@ func (s *Server) getContainerLogs(request *restful.Request, response *restful.Re
...
+			return nil
+		}
+
+		l, err := r.ReadBytes('\n')
+		if err != nil {
+			if err != io.EOF {
+				return fmt.Errorf("failed to read log file: %q: %v", logInfo.LogsFile, err)
+			}
+
+			if opts.follow {
```
wzshiming, line 261:
```suggestion
			if opts.follow && logInfo.Follow {
```

test/kwokctl/kwokctl_logs_test.sh
nikola-jokic: Sounds great to me! On it :relaxed:

test/kwokctl/kwokctl_kind_logs.test.sh
```diff
@@ -0,0 +1,30 @@
...
+DIR="$(dirname "${BASH_SOURCE[0]}")"
+
+DIR="$(realpath "${DIR}")"
+
+source "${DIR}/helper.sh"
+
+function main() {
+  local all_releases=("${@}")
+
+  test_all "kind" "logs" "${all_releases[@]}" || exit 1
```
nikola-jokic, line 25: This test will fail because `/tmp/kwok.log` is not going to be present
One solution is to read config the same way it is being done in `components.BuildKwokControllerComponent`
That way, we can extract log mounts and add them to extra mounts.

I just wanted to confirm that this is okay, or is there another way I should approach this? :relaxed:

pkg/kwokctl/runtime/kind/cluster.go
```diff
@@ -71,6 +73,38 @@ func getKindRuntimeExtraArgs(config *internalversion.KwokctlConfiguration, s str
 	return []internalversion.ExtraArgs{}
 }
 
+func getKindRuntimeExtaLogMounts(configPath string) ([]internalversion.Volume, error) {
+	objs, err := config.Load(context.Background(), configPath)
```
wzshiming, line 77:
```suggestion
func GetKindRuntimeExtaLogMounts(ctx context.Context, configPath string) ([]internalversion.Volume, error) {
	objs, err := config.Load(ctx, configPath)
```

The configuration is taken from the Context.
Let's move this function to `pkg/kwokctl/runtime` to allow docker runtime and Kind runtime to use it.

pkg/kwokctl/runtime/kind/kind.yaml.tpl
```diff
@@ -104,7 +104,11 @@ nodes:
   - hostPath: {{ .ConfigPath }}
     containerPath: /etc/kwok/kwok.yaml
     readOnly: true
-
+{{ range .ExtraLogMounts }}
```
wzshiming, line 107: #439 I will add `ExtraVolumes` for additional mount directories. Please review it if you like, it should work here.

nikola-jokic: Oh, of course! I'll this issue on my PR, but having your PR also helps a log :smile:

nikola-jokic:
I am not sure why Mac does not allow `/tmp/kwok` to be mounted as a volume
[This](https://github.com/kubernetes-sigs/kwok/actions/runs/4600827433/jobs/8127925427?pr=417) workflow run shows that `/tmp/kwok` exists, but the docker on Mac does not allow it
Ideally, I'd go with this by creating a file in `/tmp`, but then the kind would fail mounting `tmp` twice to the kind runtime.
I did not try ignoring `/tmp` mount in kind since it seems to me that we would tie implementation too much with the runtime, but that might be a way to go.
@wzshiming what do you think?

test/kwokctl/kwokctl_logs_test.sh
```diff
@@ -0,0 +1,137 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+DIR="$(dirname "${BASH_SOURCE[0]}")"
+
+DIR="$(realpath "${DIR}")"
+
+RELEASES=()
+
+LOGDIR="/tmp/kwok"
```
wzshiming, line 22:
```suggestion
LOGDIR="./logs"
```

test/kwokctl/logs.yaml
```diff
@@ -0,0 +1,26 @@
+kind: Logs
+apiVersion: kwok.x-k8s.io/v1alpha1
+metadata:
+  name: fake-pod
+  namespace: other
+spec:
+  logs:
+    - containers:
+      - fake-pod
+      logsFile: /tmp/kwok/kwok.log
```
wzshiming, line 10: If this is a relative path, then it is either relative to the configuration file itself or relative to pwd?

wzshiming (commented):
I think it's a permissions issue, we don't have access to `/tmp/` on the Mac in CI

test/kwokctl/logs.yaml
nikola-jokic: I added it to the `config.Load`
It seemed like a nice place since the defaults were applied there, and at that point, we know what is the working directory.
Now the linter is complaining about it, but to me, it seemed like the most appropriate place :smile:

nikola-jokic: But I think we are still going to bump into the same linter issue when we extend another config option like `Attach`

nikola-jokic:
Trying tests again

If they fail, I'll investigate it to see if there is a problem on the runner side... I executed logs tests on Mac, and they worked... It might be something similar to the hosted machine from the last time :confused:

nikola-jokic:
Update: I think it is something to do with colima and latest mac release: https://github.com/abiosoft/colima/issues/503
I think I found a way to fix it

.github/workflows/test.yaml
```diff
@@ -160,7 +160,7 @@ jobs:
         shell: bash
         run: |
           brew install colima docker
-          colima start --mount $HOME/.kwok/:w
+          colima start --mount $HOME/.kwok/:w --mount $(pwd):w --mount-type virtiofs
```
nikola-jokic, line 163: The reason Mac workflows were failing... There was an [issue](https://github.com/abiosoft/colima/issues/503) in colima with latest mac causing this problem

pkg/kwokctl/runtime/kind/kind.go
```diff
@@ -127,4 +127,5 @@ type BuildKindConfig struct {
 	SchedulerExtraVolumes         []internalversion.Volume
 	ControllerManagerExtraArgs    []internalversion.ExtraArgs
 	ControllerManagerExtraVolumes []internalversion.Volume
+	ExtraLogVolumes               []internalversion.Volume
```
wzshiming, line 130: Can We use `ExtraVolumes` for this instead?

https://github.com/kubernetes-sigs/kwok/blob/72824baf0362ba6c309554379185c6b180b1fb07/pkg/kwokctl/runtime/kind/kind.go#L47

test/kwokctl/kwokctl_logs_test.sh
```diff
@@ -0,0 +1,135 @@
...
+      echo " got: ${result}"
+      return 1
+  fi
+}
+
+function test_apply_node_and_pod() {
+  kwokctl --name "${name}" kubectl apply -f "${DIR}/fake-node.yaml"
+  if [[ $? -ne 0 ]]; then
+    echo "Error: fake-node apply failed"
+    exit 1
```
wzshiming, line 80:
```suggestion
    return 1
```

test/kwokctl/kwokctl_logs_test.sh
```diff
@@ -0,0 +1,135 @@
...
+  fi
+  kwokctl --name "${name}" kubectl apply -f "${DIR}/fake-pod-in-other-ns.yaml"
+  if [[ $? -ne 0 ]]; then
+    echo "Error: fake-pod apply failed"
+    exit 1
```
wzshiming, line 85:
```suggestion
    return 1
```

test/kwokctl/kwokctl_logs_test.sh
```diff
@@ -0,0 +1,135 @@
...
+  fi
+  kwokctl --name "${name}" kubectl apply -f "${DIR}/fake-deployment.yaml"
+  if [[ $? -ne 0 ]]; then
+    echo "Error: fake-deployment apply failed"
+    exit 1
```
wzshiming, line 90:
```suggestion
    return 1
```

test/kwokctl/kwokctl_logs_test.sh
```diff
@@ -0,0 +1,135 @@
...
+  fi
+  kwokctl --name "${name}" kubectl wait pod -A --all --for=condition=Ready --timeout=60s
+  if [[ $? -ne 0 ]]; then
+    echo "Error: fake-pod wait failed"
+    echo kwokctl --name "${name}" kubectl get pod -A --all
+    kwokctl --name "${name}" kubectl get pod -A --all
+    exit 1
```
wzshiming, line 97:
```suggestion
    return 1
```

pkg/kwokctl/components/kwok_controller.go
```diff
@@ -73,6 +78,34 @@ func BuildKwokControllerComponent(conf BuildKwokControllerComponentConfig) (comp
 				ReadOnly:  true,
 			},
 		)
+
+		objs, err := config.Load(context.Background(), conf.ConfigPath)
```
wzshiming, line 82:
```suggestion
		logs := config.FilterWithTypeFromContext[*internalversion.Logs](ctx)
		clusterLogs := config.FilterWithTypeFromContext[*internalversion.ClusterLogs](ctx)
```

nikola-jokic: I like this, and we can leverage `runtime.GetLogVolumes` to avoid logic replication. Then only thing is that we don't have a context here. Can I change the function signature to accept context?

wzshiming: It seems that this should be moved outside of `BuildKwokControllerComponent` so that the unit tests for `BuildKwokControllerComponent` have no dependencies, if any.

nikola-jokic: That is even better

test/kwokctl/logs.yaml
wzshiming: Yes, I think `attach` is actually `logs -f` if the input is ignored.

pkg/kwokctl/runtime/kind/cluster.go
```diff
@@ -108,10 +108,18 @@ func (c *Cluster) Install(ctx context.Context) error {
...
+
 	etcdComponentPatches := runtime.GetComponentPatches(config, "etcd")
 	kubeApiserverComponentPatches := runtime.GetComponentPatches(config, "kube-apiserver")
 	kubeSchedulerComponentPatches := runtime.GetComponentPatches(config, "kube-scheduler")
 	kubeControllerManagerComponentPatches := runtime.GetComponentPatches(config, "kube-controller-manager")
+	extraLogVolumes, err := runtime.GetLogVolumes(ctx)
+	if err != nil {
+		return err
+	}
+	kwokControllerExtraVolumes := kwokControllerComponentPatches.ExtraVolumes
```
nikola-jokic, line 121: Is there a reason why `KwokControllerComponentPatches.ExtraVolumes` are not mounted on kind? I wanted to merge them but as soon as I did, this workflow started failing: https://github.com/kubernetes-sigs/kwok/actions/runs/4631661137/jobs/8194800586?pr=417

It is failing on my machine, saying that `/tmp` mount is used multiple times (we used /tmp on all mounts)

If we should not include it, I'll fix this no problem, but I thought we are going to need extra mounts for the kwok controller as well

nikola-jokic: We can probably do a PR to allow relative paths that will fix this one and extend the extra volume mounts? I can do it here but I'm afraid it is going to be too much for one PR (It is already huge)

pkg/kwok/server/debugging_logs.go
```diff
@@ -89,3 +125,445 @@ func (s *Server) getContainerLogs(request *restful.Request, response *restful.Re
...
+	if apiOpts.SinceTime != nil && apiOpts.SinceTime.After(opts.since) {
+		opts.since = apiOpts.SinceTime.Time
+	}
+	return opts
+}
+
+func readLogs(ctx context.Context, logInfo *internalversion.Log, opts *logOptions, stdout, stderr io.Writer) error {
+	logger := log.FromContext(ctx)
+
+	logsFilePath := path.Join("/var/components/controller", logInfo.LogsFile)
```
wzshiming, line 217: I think this part of the logic should be handled in kwokctl, not in the kwok server.

wzshiming (commented):
I've tried `kubectl logs` which works fine, `kubectl logs -f` which doesn't seem to work

pkg/kwok/server/debugging_logs.go
nikola-jokic: Working on it
Thank you for reviews a lot! I wanted to push this to ask for an opinion

I'll have it ready now that I know everything how to structure it, and I'll test all scenarios
Thank you again!

pkg/kwokctl/runtime/util.go
```diff
@@ -43,3 +49,47 @@ func ExpandVolumesHostPaths(volumes []internalversion.Volume) ([]internalversion
...
+				return nil, err
+			}
+			mountDirs[filepath.Dir(abs)] = struct{}{}
+		}
+	}
+
+	volumes := make([]internalversion.Volume, 0, len(mountDirs))
+	i := 0
+	for dir := range mountDirs {
+		dirPath := strings.TrimPrefix(dir, "/var/components/controller")
```
nikola-jokic, line 83: This is the part I am unsure of. We can probably add strip prefix only in case when runtime is kind because we know that the prefix is going to be added, and it is going to be used in template.

The issue is that to use extra volume mounts, we add `/var/components/controller` to all paths.
Since we need correct configuration for logs, we expand the path. I agree that the server should not inspect the runtime from configuration, so that is why this change was necessary. Ideally it should not interfere with the binary runtime unless the actual path there is this constant. Because of that, It might be wise to use check for the runtime but I wanted to check first, since this approach might be flawed in a way I don't know :smile:

nikola-jokic:
Thank you for pointing out the follow. I included tests for tail, since and follow in e2e tests :smile:

wzshiming:
/approve
/lgtm
/hold

I've tested it locally and it's all working as expected, great!
Unhold as you wish 😄

Also, add the same logic to Attach, do you think it's appropriate?

nikola-jokic:
/unhold

Oh sure! Taking that one :smile:

PR #433, opened 2023-03-30
add unit test for selector

lixingchenDaoCloud:
#### What type of PR is this?

/kind feature

add unit test for selector

Fixes #

NONE

### DIFF
--- a/pkg/utils/expression/selector_test.go
+++ b/pkg/utils/expression/selector_test.go
@@ -1,0 +1,150 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


pkg/utils/expression/selector_test.go
```diff
@@ -0,0 +1,150 @@
+/*
+Copyright 2022 The Kubernetes Authors.
```
wzshiming, line 2:
```suggestion
Copyright 2023 The Kubernetes Authors.
```

lixingchenDaoCloud: Thanks for your review ,and i have change this problem

wzshiming (approved):
/approve
/lgtm

PR #457, opened 2023-04-07
snapshot: add limitrange, priorityclass and runtimeclass to default resources

Huang-Wei:
#### What type of PR is this?

/kind cleanup
/kind documentation

In regular prod clusters, usually namespace is associated with LimitRange, and Pods are associated with RuntimeClass and PriorityClass. Not scraping them by default would cause the snapshot infeasible to be imported/restored.

Also this PR explains how a resource string is parsed to a GVR object.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kwokctl/snapshot/runtime.go
+++ b/pkg/kwokctl/snapshot/runtime.go
@@ -25,12 +25,19 @@ import (
 )
 
 // Resources is the resources of cluster want to save or restore
+// The resource string format adheres to a GVR template.
+// - a single string like "node" will be parsed as {resource: node}
+// - a two-section string like "daemonset.apps" will be parsed as {resource: dasemonset, group: apps}
+// - a three-section (or 3+) string like "foo.v1alpha1.example.com" will be parsed as {resource: foo, version: v1alpha1, group: example.com}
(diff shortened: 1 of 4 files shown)


pkg/kwokctl/snapshot/runtime.go
```diff
@@ -25,12 +25,19 @@ import (
...
+// - a three-section (or 3+) string like "foo.v1alpha1.example.com" will be parsed as {resource: foo, version: v1alpha1, group: example.com}
 var Resources = []string{
 	"namespace",
 	"node",
 	"serviceaccount",
 	"configmap",
 	"secret",
+	"limitrange",
+	"runtimeclass.v1.node.k8s.io",
+	"priorityclass.v1.scheduling.k8s.io",
```
wzshiming, line 40: We can remove v1, which will allow it to be adapted to the current version actually supported by the cluster

```suggestion
	"runtimeclass.node.k8s.io",
	"priorityclass.scheduling.k8s.io",
```

Huang-Wei: Updated. PTAL.

PR #458, opened 2023-04-07
snapshot: impose retry logic

Huang-Wei:
#### What type of PR is this?

/kind feature

Impose retry logic to snapshotting.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kwokctl/snapshot/save.go
+++ b/pkg/kwokctl/snapshot/save.go
@@ -21,11 +21,14 @@ import (
 	"fmt"
 	"io"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/tools/pager"
+	"k8s.io/client-go/util/retry"
 
 	"sigs.k8s.io/kwok/pkg/utils/client"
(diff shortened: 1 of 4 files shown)


pkg/kwokctl/snapshot/save.go
```diff
@@ -62,23 +65,31 @@ func Save(ctx context.Context, kubeconfigPath string, w io.Writer, resources []s
...
+			listPager := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
+				return nri.List(ctx, opts)
+			})
+			return listPager.EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
+				if o, ok := obj.(metav1.Object); ok {
+					clearUnstructured(o)
+				}
+				return encoder.Encode(obj)
+			})
+		}); err != nil {
```
wzshiming, line 80: We could have just added retries to the pagination list, which might have worked better.

```suggestion
		listPager := pager.New(func(ctx context.Context, opts metav1.ListOptions) (list runtime.Object, err error) {
			err = retry.OnError(retry.DefaultBackoff, retriable, func() error {
				list, err = nri.List(ctx, opts)
				return err
			})
			return list, err
		})
		err = listPager.EachListItem(ctx, metav1.ListOptions{}, func(obj runtime.Object) error {
			if o, ok := obj.(metav1.Object); ok {
				clearUnstructured(o)
			}
			return encoder.Encode(obj)
		})
		if err != nil {
```

Huang-Wei: +1, we should catch and retry the underneath API call, in which case is the [pageFn](https://github.com/kubernetes/kubernetes/blob/ad18954259eae3db51bac2274ed4ca7304b923c4/staging/src/k8s.io/client-go/tools/pager/pager.go#L237) in the `pager` object.

Updated.

PR #495, opened 2023-04-14
fix waitForReady no error return

helen-frank:
#### What type of PR is this?
/kind bug

waitForReady unable to return error

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kwok/cmd/root.go
+++ b/pkg/kwok/cmd/root.go
@@ -236,22 +236,17 @@ func waitForReady(ctx context.Context, clientset kubernetes.Interface) error {
 		Steps:    5,
 	}
 
-	err := wait.Poll(ctx,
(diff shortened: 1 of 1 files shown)


wzshiming (commented):
This is expected, only the error message is output instead of the immediate end of the poll.

helen-frank:
> This is expected, only the error message is output instead of the immediate end of the poll.

Is it because kwok does not have node at runtime?

wzshiming:
Just waiting for the apiserver to be ready.

helen-frank:
> Just waiting for the apiserver to be ready.

An idea, delete this code

wzshiming:
This is to avoid the controller constantly restarting when the apiserver is not ready

helen-frank:
got it and /close

PR #555, opened 2023-04-28
[kwok] Provide generic functions that enable it to simulate data for any indicator (Node)

nikola-jokic:
#### What type of PR is this?

/kind feature

Related to #199


nikola-jokic:
This is currently just a proposed way of exposing metrics from the controller, to make sure package structure is okay.
Looking forward to your suggestions @wzshiming :smile:

wzshiming:
The package structure is good to me,
I would like to minimise hard-coded metrics and use configuration whenever possible 😄

pkg/apis/internalversion/metric_type.go
```diff
@@ -0,0 +1,72 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Metric provides metrics configuration.
+type Metric struct {
+	//+k8s:conversion-gen=false
+	metav1.TypeMeta `json:",inline"`
+	// Standard list metadata.
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	metav1.ObjectMeta `json:"metadata"`
```
nikola-jokic, line 29: Using `json` here in order to execute template correctly like expected :relaxed: 
The expression `{{ .metadata.name }}` would not be evaluated correctly without it, since the `metav1.ObjectMeta` is embedded

pkg/kwok/server/metrics.go
```diff
@@ -17,11 +17,48 @@ limitations under the License.
...
-	promHandler := promhttp.Handler()
-	s.restfulCont.Handle("/metrics", promHandler)
+	defaultMetrics := metrics.DefaultMetrics()
+	s.restfulCont.Handle("/metrics", defaultMetrics.Handler())
+	for _, m := range s.config.Metrics {
+		path, err := executePathTemplate(m.Spec.Path, m)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to execute path template: %v", err))
+		}
+		reg := metrics.CustomMetric(m.Name)
```
nikola-jokic, line 38: The purpose of this is just to create an empty registry and install handler. The registry metric installation is lazy. It will install metrics as soon as the `node` comes up in order to evaluate labels correctly :relaxed:

test/kwokctl/metric.yaml
```diff
@@ -0,0 +1,49 @@
...
+  # I'm not sure if there is a way to configure the `metric-server` to change the path of the node metrics collection
+  path: "/metrics/nodes/{{ .metadata.name }}"
+  metrics:
+    - name: kubelet_node_name
+      help: "[ALPHA] The node's name. The count is always 1."
+      kind: gauge
+      labels:
+        - name: node
+          value: '{{ .metadata.name }}'
+      value: '1.0'
```
nikola-jokic, line 15: Used 1.0 here, since the `cel` would evaluate it to an integer. Another approach I can take is to check the type and convert it. Not sure which one to take since the problem also exists on `value` field when you want to divide... Basically, you either specify `10.0` for example, or `double(10)`

pkg/kwok/controllers/node_controller.go
```diff
@@ -277,6 +284,7 @@ func (c *NodeController) watchResources(ctx context.Context, opt metav1.ListOpti
 					node := event.Object.(*corev1.Node)
 					if c.need(node) {
 						c.putNodeInfo(node)
+						c.applyMetrics(ctx, node)
```
nikola-jokic, line 287: Should we delete the registry when the node is deleted, or should we set values to 0?

nikola-jokic:
`startedContainersTotal` is kind of tricky. The fastest approach is to do it on the Pod Controller. Basically, this function would evaluate to true if `node.metadata.name` is pod's `NodeName`, and then increase counter on `true`.
Is there a better way I'm not seeing?

pkg/kwok/metrics/cel/evaluate.go
```diff
@@ -0,0 +1,84 @@
...
+	Evaluate(exp string) (float64, error)
+}
+
+// NewNodeEvaluator returns a MetricEvaluator that is able to evaluate node metrics
+func NewNodeEvaluator(node *corev1.Node) (Evaluator, error) {
+	b, err := json.Marshal(node)
+	if err != nil {
+		return nil, err
+	}
+	m := map[string]any{}
```
nikola-jokic, line 41: Is there a better way I can leverage proto buffers instead of json? I had problems doing it, so I just wanted to have something, and then improve it later :smile:

pkg/kwok/server/metrics.go
```diff
@@ -17,11 +17,49 @@ limitations under the License.
...
-func (s *Server) InstallMetrics() {
-	promHandler := promhttp.Handler()
-	s.restfulCont.Handle("/metrics", promHandler)
+func (s *Server) InstallMetrics() error {
+	defaultMetrics := metrics.DefaultMetrics()
+	s.restfulCont.Handle("/metrics", defaultMetrics.Handler())
+	for _, m := range s.config.Metrics {
+		path, err := executePathTemplate(m.Spec.Path, m)
+		if err != nil {
+			return err
```
nikola-jokic, line 36: What should we do in case path template is incorrect? Log the warning and continue, or stop the installation?

pkg/kwok/controllers/metrics.go
```diff
@@ -0,0 +1,146 @@
...
+	// state
+	nodeContainersCounter maps.SyncMap[string, float64]
+	nodeCache             maps.SyncMap[string, *corev1.Node]
+
+	// configuration
+	metrics []*internalversion.Metric
+}
+
+func (h *nodeMetricHandler) updateCache(node *corev1.Node) {
+	h.nodeCache.Store(node.Name, node)
```
nikola-jokic, line 49: Since metric updates can occur from the `NodeController` and the `PodController`, I thought it would be best to cache the latest node state and apply all metrics on publish. I wanted to avoid using a client to fetch the node in case `PodController` updates the number of containers running on a `Node`.
Does it make sense?

pkg/kwok/metrics/cel/evaluate.go
```diff
@@ -0,0 +1,88 @@
...
+	if err := json.Unmarshal(b, &m); err != nil {
+		return nil, err
+	}
+
+	env, err := cel.NewEnv(
+		cel.Lib(&timeLib{}),
+		cel.Lib(&nodeLib{startedContainersTotal: startedContainersTotal}),
+		cel.Declarations(decls.NewVar("node", decls.NewMapType(decls.String, decls.Dyn))),
```
wzshiming, line 49: The use of CEL seems rather cumbersome, I'll try to refactor it.

wzshiming (commented):
That's great, I've run it and it looks good

pkg/kwok/controllers/metrics.go
```diff
@@ -0,0 +1,146 @@
...
+	"sigs.k8s.io/kwok/pkg/kwok/metrics/cel"
+	"sigs.k8s.io/kwok/pkg/utils/maps"
+	"sigs.k8s.io/kwok/pkg/utils/slices"
+)
+
+var metricsHandler nodeMetricHandler
+
+func setMetrics(metrics []*internalversion.Metric) {
+	metricsHandler.metrics = metrics
+}
```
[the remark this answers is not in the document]
nikola-jokic: Seems appropriate :relaxed:
Since PodController is a public package level variable, I wanted to still have a way of updating the node metric like total number of started containers. If someone chooses not to use Controller, but to instantiate PodController, then they would not have a way of updating metrics. In that case, the metricHandler would need to be exposed as well, and I did not want to extend the API surface with this :relaxed: 
Is that a valid concern at all? Maybe metric handler can go into `sigs.k8s.io/kwok/pkg/kwok/controllers/metrics` and have their methods public?

pkg/apis/internalversion/metric_type.go
```diff
@@ -0,0 +1,72 @@
...
+package internalversion
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Metric provides metrics configuration.
+type Metric struct {
+	//+k8s:conversion-gen=false
+	metav1.TypeMeta `json:",inline"`
```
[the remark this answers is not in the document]
wzshiming: The `metadata` is not in

```
metav1.TypeMeta `json:",inline"`
```

that is in 

```
metav1.ObjectMeta `json:"metadata"`
```

Is it right?

pkg/apis/internalversion/metric_type.go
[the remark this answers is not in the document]
wzshiming: It looks like this tag needs to be retained.

pkg/apis/internalversion/metric_type.go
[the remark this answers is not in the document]
nikola-jokic: Exactly! It will be embedded on default serialization like:
`{"name": "fake-node", ...}`, but we really want `{"metadata":{"name":"fake-node"}}`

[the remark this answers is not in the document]
wzshiming: Ohh, I see, our understanding of this is a little off,

I meant that `{{ .metadata.name }}` is filled with `corev1.Node` and not `internalversion.Metric`.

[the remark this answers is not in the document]
nikola-jokic: Oh, I see, but the problem persists for the `path`:
```yaml
path: "/metrics/nodes/{{ .metadata.name }}"
```
My understanding was to create path by executing a template. At this point, we don't have a node yet. We can eliminate the need for `json` if the path is not dynamic like this, but rather a constant. That way, we can always use `corev1.Node` which does have `json` struct tags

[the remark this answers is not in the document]
nikola-jokic: Oh then I misunderstood the spec for:
```yaml
name: node-name
```
So the name of the metric has nothing to do with the node it targets? If that is the case, I can change the server implementation to search for the `{{ }}` in the URL and treat it as a dynamic parameter (replace it with `/metrics/nodes/{node-name}`). It is just going to be slightly more complicated if there are multiple field selectors like `{{ .metadata.namespace }}`. Then this information would need to be propagated to the `CustomRegistry` so it knows how to fetch the correct registry for a specific endpoint. I just need to figure out how to propagate that metadata to the registry, because the metrics updates are happening per node, while the selector can have a brother scope (in this case, the namespace)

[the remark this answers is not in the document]
nikola-jokic: Great :smile: on it!

test/kwokctl/metric.yaml
wzshiming: The evaluation values can be converted in this way

``` golang
	switch v := refVal.(type) {
	case types.Duration:
		return float64(v.Duration), nil
	case types.Int:
		return float64(v), nil
	case types.Double:
		return float64(v), nil
	case types.Uint:
		return float64(v), nil
	case types.Bool:
		if v {
			return 1, nil
		}
		return 0, nil
	default:
		return 0, fmt.Errorf("unsupported metric value type: %T", v)
	}
```

test/kwokctl/metric.yaml
```diff
@@ -0,0 +1,49 @@
+kind: Metric
+apiVersion: kwok.x-k8s.io/v1alpha1
+metadata:
+  name: fake-node
+spec:
+  # I'm not sure if there is a way to configure the `metric-server` to change the path of the node metrics collection
+  path: "/metrics/nodes"
```
wzshiming, line 7:
```suggestion
  path: "/metrics/nodes/fake-node"
```

test/kwokctl/kwokctl_metric_test.sh
```diff
@@ -0,0 +1,134 @@
...
+
+function metric_value() {
+    local result="${1}"
+    local metric_name="${2}"
+    local value=$(echo "${result}" | grep "^${metric_name}" | cut -d " " -f 2)
+    echo "${value}"
+}
+
+function test_node_metrics() {
+  local result=$(curl http://localhost:10250/metrics/nodes)
```
wzshiming, line 74:
```suggestion
  local result=$(curl http://localhost:10250/metrics/nodes/fake-node)
```

pkg/kwok/controllers/pod_controller.go
```diff
@@ -420,6 +424,12 @@ func (c *PodController) watchResources(ctx context.Context, opt metav1.ListOptio
 				case watch.Added, watch.Modified:
 					pod := event.Object.(*corev1.Pod)
 					if c.need(pod) {
+						if event.Type == watch.Added && c.nodeMetricsHandler != nil {
+							c.nodeMetricsHandler.AddContainersCount(pod.Spec.NodeName, len(pod.Spec.Containers))
```
wzshiming, line 428:
```suggestion
							c.nodeMetricsHandler.AddContainersCount(pod.Spec.NodeName, 1)
```

nikola-jokic: Wouldn't that count the number of pods, not the number of containers?

pkg/kwok/controllers/pod_controller.go
```diff
@@ -420,6 +424,12 @@ func (c *PodController) watchResources(ctx context.Context, opt metav1.ListOptio
...
+							if err := c.nodeMetricsHandler.Publish(pod.Spec.NodeName); err != nil {
```
wzshiming, line 429: The Metric should not be updated when the value changes, but when it is requested

pkg/kwok/controllers/pod_controller.go
wzshiming: Yep, this is cumulative number of containers started

nikola-jokic: But if we replace len(pod.Spec.Containers) with 1, and the pod contains 5 containers for example, this would increase the count by 1, while the number of containers is actually 5? Or I completely misunderstood the metric we are trying to apply :smile:

wzshiming:
``` yaml
help: [ALPHA] Cumulative number of containers started
```

Yes this is correct, only after the kwok-controller is restarted will it start accumulating from 0 again

nikola-jokic: But changing the value from len to 1 would not really solve the restart problem since number of containers is always >= 1. I mean I or you can apply this suggestion if you think this is a right approach, I'm just trying to understand why is the len of pod.Spec.Containers wrong here :relaxed:

pkg/kwok/controllers/pod_controller.go
nikola-jokic: My understanding is that we need to update collectors so when the GET request is issued, it can pick the current state and serve it. When I looked into a kubelet code, it is basically doing the same thing:
https://github.com/kubernetes/kubernetes/blob/61d455ed1173cd89a98442adf4623a29c5681c58/pkg/kubelet/kuberuntime/kuberuntime_manager.go#L1086

We just update collectors and let the prometheus http handler serve metrics. If we evaluate it on the http request, we would need to wrap the handler to execute evaluations which will increase latency of the GET request. Publish was meant to publish metrics to collectors, and is not dependent on the scraping interval at all. As soon as the state changes, it will update the metric collectors, and that is it.

Can you please provide an example where the handler evaluates metrics and serves them? I'm curious how it works :relaxed:

pkg/kwok/controllers/metrics/metrics.go
```diff
@@ -0,0 +1,136 @@
...
+				if err != nil {
+					errs = append(errs, fmt.Errorf("failed to get/create counter for metric %q with configuration name %q: %w", metric.Name, m.Name, err))
+					continue
+				}
+				val, err := eval.Evaluate(m.Value)
+				if err != nil {
+					errs = append(errs, fmt.Errorf("failed to evaluate expression %q in metric %q: %w", m.Value, metric.Name, err))
+					continue
+				}
+				c.Add(val)
```
wzshiming, line 106: https://github.com/kubernetes-sigs/kwok/pull/555#discussion_r1202237446

I've found it a bit tricky, as it causes each update to increase the current total number of containers.

nikola-jokic: Oh but it should be 2 right? Because kubelet started 2 containers. One once you created a pod, and another one once you modified it, it re-created a container. My understanding of this metric is to say how many containers did the kubelet start, not how many of them are running at this time

pkg/kwok/controllers/pod_controller.go
nikola-jokic: This is awesome :100: ! I just updated the code, and thank you so much for this one!

pkg/kwok/controllers/metrics/metrics.go
wzshiming: Your understanding is correct.

But the `val` value is the cumulative total of containers started.

Each execution of `c.Add(val)` will cause the value in the indicator to increase the cumulative number of containers started.

nikola-jokic: I have been thinking how can we implement this with the ContainerFunc and I don't know how effectively to solve it.
I had another idea that is possibly not that good (I don't know nearly this as much as you do). Can we leverage stages? 

Since the pod controller keeps track of the resource version and it computes the stage transition, can we leverage that? When the pod controller is about to set the stage to `ContainersReady`, we increment the count. That way, we can avoid allocating memory and trying to figure out if this container is about to be executed (there is already logic there that keeps track of the stage transitions). And also, we don't need to react on events. As soon as the controller knows it should transition the pod to `ContainersReady`, that is when we increment the count. The `Add` event can be then avoided (cache re-sync), and the metric would be correct (what if the stage applied is to simulate a faulty container image that couldn't be pulled from the container registry, for example. Then the number of container started should not be incremented).
What do you think?

wzshiming: 🤣I think it should work. I will give it a try.

nikola-jokic: Found a problem
THe value: `( unixSecond(now()) - unixSecond(node.metadata.creationTimestamp) ) / 10.0` can't work since the creationTimestamp is a string, and unixSeconds accepts Time. In order to be able to work with this interface, should we change now() to return a string? Then unixSecond will accept that, parse it as time and return seconds.
We can also introduce another function called time for example, to parse string to a time

What would you prefer?

wzshiming: https://github.com/kubernetes-sigs/kwok/commit/3cb4ae98ac14279f52351e3f914ea321bb91c095

I've refactored, which should solve both of your problems. 😄
But I've found a new problem with the Histogram implementation, it doesn't work as expected, and I think the configuration needs to be changed.

nikola-jokic: Did I write over it with rebase :disappointed: ? I'll see the diff and try to apply it back, sorry :smile:

nikola-jokic: I see some comments, are you refactoring this now? It would be awesome if you push it to this branch :relaxed:

wzshiming: done 😄

pkg/kwok/metrics/metrics.go
```diff
@@ -0,0 +1,215 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to register gauge %q: %w", m.Name, err)
+			}
+			eval, err := h.environment.Compile(m.Value)
+			if err != nil {
+				return fmt.Errorf("failed to compile metric %q: %w", metric.Name, err)
+			}
+
+			updateFunc := func() error {
+				nodeInfo, ok := h.getNodeInfo(metric.Name)
```
nikola-jokic, line 109: Doesn't this mean that the `internalversion.Metric` corresponds to a single `corev1.Node`? 
Follow-up question: to handle labels, we need to register metrics only when the node is in the cache. 
How do you feel about using closure there with sync.Once? Once the `updateFunc` is nil, it won't have a reference to a closure so the closure will be garbage collected, and we don't need to extend any struct
On each call to updateFunc, we call once.Do(register metric with node), and then evaluate a program
We can't really register metric on `NewMetricUpdateHandler` since the metric will be registered before the node is created, and we need the node in order to apply constant labels :relaxed:

pkg/kwok/metrics/metrics.go
```diff
@@ -0,0 +1,340 @@
...
+
+				var err error
+				once.Do(func() {
+					err = registerFunc(nodeInfo.Node)
+				})
+				if err != nil {
+					return err
+				}
+
+				for le, eval := range bucketEvals {
```
nikola-jokic, line 272: Is this the logic that should be applied? Once this is done, I think it would be good for the complete review
Thank you for improving this PR @wzshiming !

pkg/kwok/metrics/histogram_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"github.com/gogo/protobuf/proto"
+	dto "github.com/prometheus/client_model/go"
+)
+
+func TestHistogramObserve(t *testing.T) {
```
nikola-jokic, line 11: @wzshiming structured this test as a POC, we can change it any time
Is this what you have meant? Any pointers are helpful :smile:

wzshiming: Thank you, 😄
IMO If this value is cumulative, it will increase in proportion to the rate at which metrics are collected. 
So we should give the value calculated to `Set` instead of `Inc`.

nikola-jokic: Oh I agree 100%, great suggestion! I'll try refactoring a bit, to make it more readable
I have noticed some resets in prometheus.Histogram, I don't know how or when is it done, so I'll try digging into it
Maybe they are just internal for the purpose of memory and performance. I also tried not using bit manipulations to increase performance. This is more readable and easier to debug and understand

pkg/kwok/metrics/histogram.go
```diff
@@ -0,0 +1,181 @@
...
+	ch <- h.desc
+}
+
+// Collect sends histogram to a prometheus Metric channel.
+func (h *Histogram) Collect(ch chan<- prometheus.Metric) {
+	ch <- h
+}
+
+// Set is used to increment value at le bucket and add val to the sum.
+func (h *Histogram) Set(le string, val float64) (err error) {
```
wzshiming, line 153: I tested it and it didn't behave as I expected, I refactored it and now it meets expectations 😄.

https://github.com/wzshiming/kwok/commit/e4f49e3acf3ae0915cb3898e8f9cbf326888aa93

`git apply <(curl https://github.com/wzshiming/kwok/commit/e4f49e3acf3ae0915cb3898e8f9cbf326888aa93.patch)`

nikola-jokic: Amazing, sorry I could not make it work right away, I did not exactly know how it should behave :smile: 
I tried applying patch but it produced an error
Do you mind pushing on the branch :smile:

wzshiming: BTY, I added [auto docs for API](https://kwok.sigs.k8s.io/docs/generated/apis/), so this PR needs to rebase then exec `./hack/update-api-docs.sh`

nikola-jokic: Rebased and executed
I'll apply the patch by hand. git shows error (patch does not apply)

pkg/apis/v1alpha1/metric_types.go
```diff
@@ -0,0 +1,82 @@
...
+	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
+	metav1.ObjectMeta `json:"metadata"`
+	// Spec holds spec for metrics.
+	Spec MetricSpec `json:"spec"`
+}
+
+// MetricSpec holds spec for metrics.
+type MetricSpec struct {
+	// Path is a restful service path.
+	Path string `json:"path"`
```
wzshiming, line 44: Suggestion, if kwokctl is using this configuration, we should add this Path to the Prometheus collection target, so we can open the web of Prometheus and see the chart of the configured metrics.

WDYT?

test/kwokctl/metric.yaml
```diff
@@ -0,0 +1,54 @@
+kind: Metric
+apiVersion: kwok.x-k8s.io/v1alpha1
+metadata:
+  name: fake-node
+spec:
+  # I'm not sure if there is a way to configure the `metric-server` to change the path of the node metrics collection
```
wzshiming, line 6:
```suggestion
```

Remove this comments

pkg/kwok/metrics/cel/evaluate.go
```diff
@@ -0,0 +1,211 @@
...
+	return evaluator, nil
+}
+
+// Evaluator evaluates a cel program
+type Evaluator struct {
+	program cel.Program
+}
+
+// EvaluateMetric evaluates a cel program and returns a metric value
+func (e *Evaluator) EvaluateMetric(node *corev1.Node) (float64, error) {
```
wzshiming, line 170:
```suggestion
func (e *Evaluator) EvaluateFloat64(node *corev1.Node) (float64, error) {
```

Consistent with the style of `EvaluateString`?

hack/tools/tools.go
```diff
@@ -22,9 +22,8 @@ limitations under the License.
 package tools
 
 import (
-	// code-generator
-	_ "k8s.io/code-generator"
-
 	// gen-crd-api-reference-docs
 	_ "github.com/ahmetb/gen-crd-api-reference-docs"
+	// code-generator
```
wzshiming, line 27: Revert this file

nikola-jokic: Oh sure, sorry about that

pkg/apis/v1alpha1/metric_types.go
nikola-jokic: That sounds great as long as we keep it static
If you see it become dynamic (having route like `/nodes/{nodeName}`) in the future, is it okay to opt-out of the prometheus collection targets?

wzshiming: Prometheus supports rules for dynamically discovered targets, which will be designed in the future.

test/kwokctl/metric.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - name: kubelet_node_name
+      help: "[ALPHA] The node's name. The count is always 1."
+      kind: gauge
+      labels:
+        - name: node
+          value: 'node.metadata.name'
+      value: '1.0'
+    - name: kubelet_started_containers_total
+      # For special cases, we can also consider providing special functions to provide data,
+      # such as this number of containers started.
```
nikola-jokic, line 18:
```suggestion
```

test/kwokctl/metric.yaml
```diff
@@ -0,0 +1,51 @@
...
+      kind: gauge
+      labels:
+        - name: node
+          value: 'node.metadata.name'
+      value: '1.0'
+    - name: kubelet_started_containers_total
+      help: "[ALPHA] Cumulative number of containers started"
+      kind: counter
+      value: 'startedContainersTotal( node.metadata.name )'
+    - name: kubelet_pleg_relist_duration_seconds
```
wzshiming, line 19: Look nice 😄

image

pkg/kwok/controllers/pod_controller.go
```diff
@@ -507,6 +507,13 @@ func (c *PodController) watchResources(ctx context.Context, opt metav1.ListOptio
 						)
 					}
 
+					if event.Type == watch.Added &&
```
wzshiming, line 510: We should stop collecting this data when no Metrics

hack/tools/tools.go
```diff
@@ -28,3 +28,5 @@ import (
 	// gen-crd-api-reference-docs
 	_ "github.com/ahmetb/gen-crd-api-reference-docs"
 )
+
+
```
wzshiming, line 32:
```suggestion

```

https://github.com/kubernetes-sigs/kwok/blob/6d51ec0f44152a4b13e532a416ecdaca45434ef5/hack/verify-go-format.sh#L24

It looks like format's check didn't cover this file 😓

wzshiming:
BTY, commits are too much, squash it 😄

pkg/kwok/controllers/controller.go
```diff
@@ -194,6 +195,15 @@ func NewController(conf Config) (*Controller, error) {
 		return nil, fmt.Errorf("failed to create nodes controller: %w", err)
 	}
 
+	nodeHasMetric := func(nodeName string) bool {
+		for _, m := range conf.Metrics {
+			if m.Name == nodeName {
+				return true
+			}
+		}
+		return false
```
wzshiming, line 204:
```suggestion
		return len(conf.Metrics) != 0
```

wzshiming (approved):
/approve
/lgtm

Thank you for your contribution!

nikola-jokic:
Thank you for doing many reviews and improvements!

PR #557, opened 2023-05-02
speedup ci test-kwokctl (macos-latest, kind)

### LINKED ISSUE #536: Reduce test time "test-kwokctl (macos-latest, kind)"
### Which jobs are failing?

NONE

### Which tests are failing?

This test takes too long try to reduce it

### Since when has it been failing?

NONE

### Reason for failure (if possible)

_No response_

_No response_

Eayne:
#### What type of PR is this?
/kind cleanup

On MacOS, when http_proxy or https_proxy is explicitly set, Kubelet will be unable to access api server of locally built cluster, whose address is recorded as 0.0.0.0:port in kubeconfig created by Kind. In hence it takes so much time to wait for cluster's readiness, which exec Kubelet commands to check.

Fixes #536 

```release-note

```

```docs

```


.github/workflows/test.yaml
```diff
@@ -125,8 +125,8 @@ jobs:
       fail-fast: false
       matrix:
         os:
-          - ubuntu-latest
           - macos-latest
+          - ubuntu-latest
```
Huang-Wei, line 129: given you specified `no_proxy`, is this change still needed?

Huang-Wei:
@Eayne may i know why you close this? Is that this PR is not working as expected?

Eayne:
> @Eayne may i know why you close this? Is that this PR is not working as expected?

Yes. Usually the  test-kwokctl (macos-latest, kind) takes 60min to 80min to be finished.  When I use no_proxy, at first it can only cost 40 min to finish. As more tests being runned, I think this pr may not work since it still needs too much time. Probably it's not caused by this reason. Maybe I need to find out more to solve this issue

Eayne:
> Thanks for the efforts in investigation still 🙏

Thanks for your review also.

PR #592, opened 2023-05-15
fix verify-ends-newline

Zhuzhenghao:
#### What type of PR is this?

/kind bug

This PR adds the option to exclude files in the vendor directory when using the check_ends function. 

This ensures that files in the vendor directory are not mistakenly included in the check for newline characters at the end of files. 

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/hack/verify-ends-newline.sh
+++ b/hack/verify-ends-newline.sh
@@ -28,6 +28,12 @@ function check_ends() {
     -o -iname "*.yaml" \
     -o -iname "*.yml" \
     \) \
+    -not \( \
+    -path ./.git/\* -o \
+    -path ./vendor/\* -o \
+    -path ./demo/node_modules/\* -o \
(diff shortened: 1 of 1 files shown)


hack/verify-ends-newline.sh
```diff
@@ -28,6 +28,7 @@ function check_ends() {
     -o -iname "*.yaml" \
     -o -iname "*.yml" \
     \) \
+    -not -path ./vendor\* \
```
wzshiming, line 31: https://github.com/kubernetes-sigs/kwok/blob/df4284531ca1acfdbebe07dde574adcffbcb3772/hack/update-spelling.sh#L46-L51

Follow this and do the same for other scripts 😄

wzshiming (approved):
/approve
/lgtm

PR #619, opened 2023-05-29
feat: support `--page-size` and `--page-buffer-size` for snapshot export and save

AhmedGrati:
#### What type of PR is this?
/kind feature

This PR adds the support of `-page-size` and `-page-buffer-size` when executing `kowkctl snapshot export` 

Previously, the list pager had the default page size of `500` and `10` as the default page buffer size. Therefore, this PR aims to introduce the tuning of these parameters to users, so they can adjust them accordingly to their needs, which can increase the performance in some cases.

Fixes #

```release-note
Introduce flags `--page-size` and `--page-buffer-size` for `kowkctl snapshot export`.
```

```docs

```

### DIFF
--- a/pkg/kwokctl/runtime/cluster_snapshot.go
+++ b/pkg/kwokctl/runtime/cluster_snapshot.go
@@ -27,7 +27,7 @@ import (
 )
 
 // SnapshotSaveWithYAML save the snapshot of cluster
-func (c *Cluster) SnapshotSaveWithYAML(ctx context.Context, path string, filters []string) error {
+func (c *Cluster) SnapshotSaveWithYAML(ctx context.Context, path string, filters []string, pageSize int64, pageBufferSize int32) error {
 	file, err := os.Create(path)
 	if err != nil {
 		return err
@@ -38,7 +38,7 @@ func (c *Cluster) SnapshotSaveWithYAML(ctx context.Context, path string, filters
 	kubeconfigPath := c.GetWorkdirPath(InHostKubeconfigName)
 	// In most cases, the user should have full privileges on the clusters created by kwokctl,
 	// so no need to expose impersonation args to "snapshot save" command.
-	return snapshot.Save(ctx, kubeconfigPath, file, filters, rest.ImpersonationConfig{})
+	return snapshot.Save(ctx, kubeconfigPath, file, filters, rest.ImpersonationConfig{}, pageSize, pageBufferSize)
 }
 
 // SnapshotRestoreWithYAML restore the snapshot of cluster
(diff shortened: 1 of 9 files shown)


wzshiming:
It looks like the CI cluster is abnormal

wzshiming:
Yep, It is.

https://github.com/kubernetes/test-infra/issues/29622

pkg/kwokctl/snapshot/save.go
```diff
@@ -37,7 +37,7 @@ import (
 )
 
 // Save saves the snapshot of cluster
-func Save(ctx context.Context, kubeconfigPath string, w io.Writer, resources []string, impersonateConfig rest.ImpersonationConfig) error {
```
wzshiming, line 40: Do we define a structure to put all these extra parameters in?

pkg/kwokctl/runtime/cluster_snapshot.go
```diff
@@ -38,7 +38,7 @@ func (c *Cluster) SnapshotSaveWithYAML(ctx context.Context, path string, filters
 	kubeconfigPath := c.GetWorkdirPath(InHostKubeconfigName)
 	// In most cases, the user should have full privileges on the clusters created by kwokctl,
 	// so no need to expose impersonation args to "snapshot save" command.
-	return snapshot.Save(ctx, kubeconfigPath, file, filters, rest.ImpersonationConfig{})
+	return snapshot.Save(ctx, kubeconfigPath, file, filters, rest.ImpersonationConfig{}, pageSize, pageBufferSize)
```
wzshiming, line 41: Ohh, I seem to have misremembered 😄, it should be like `--as` and `--as-group`.

`--page-size` and `--page-buffer-size` are fixed defaults for `save`, while `export` can be changed

pkg/kwokctl/snapshot/save.go
AhmedGrati: Do you mean the pager configuration or all the parameters?

wzshiming: I mean the `rest.ImpersonationConfig` and the pager configuration

pkg/kwokctl/snapshot/save.go
```diff
@@ -89,6 +96,11 @@ func Save(ctx context.Context, kubeconfigPath string, w io.Writer, resources []s
 			return list, err
 		})
 
+		if pagerConfig != nil {
+			listPager.PageSize = pagerConfig.PageSize
+			listPager.PageBufferSize = pagerConfig.PageBufferSize
```
wzshiming, line 101: If the value is zero, then do nothing.

wzshiming (approved):
/approve
/lgtm

Huang-Wei:
/lgtm still

Thanks.

PR #693, opened 2023-06-29
[kwokctl] Add --dry-run

wzshiming:
#### What type of PR is this?

/kind feature

Fixes #

```release-note
[kwokctl] Add --dry-run
```

```docs
NONE
```


pkg/utils/dryrun/dryrun.go
```diff
@@ -0,0 +1,141 @@
...
+	"sigs.k8s.io/kwok/pkg/utils/slices"
+)
+
+// DryRun is a flag to indicate whether the program is running in dry-run mode.
+var DryRun bool
+
+// IsDryRun returns true if the program is running in dry-run mode.
+func IsDryRun() bool {
+	return DryRun
+}
```
lianghao208, line 38: Should we move these to pkg/config/vars.go? Considering dryrun is also a config var of kwokctl like others.

wzshiming: emmm, It should not be suitable, `pkg/config/vars.go` is for `kwok`/`kwokctl`, `dryrun` is only for `kwokctl` and contains some additional output functions that do not correspond to the config.

lianghao208: I mean only move the DryRun variable definition and getter function to the config package.  Putting these additional output functions in an independent util package looks good to me. IMO, it's a bit odd to set and get a config value from a util package.

wzshiming: > it's a bit odd to set and get a config value from a util package.
That makes sense, then I think we can move dryrun,file,exec to `pkg/kwokctl`.

PR #751, opened 2023-08-02
pkg/kwok/controllers: should not add duplicate job

fuweid:
#### What type of PR is this?

/kind bug

When kube-apiserver restarts, the watch might return ADD event which we already received.
We should not add the cronjob for the same node. Otherwise, the number of lease updates will be increased 2-3 times.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/kustomize/rbac/role.yaml
+++ b/kustomize/rbac/role.yaml
@@ -7,6 +7,18 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - events
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - nodes
   verbs:
   - get
--- a/pkg/kwok/controllers/node_controller.go
+++ b/pkg/kwok/controllers/node_controller.go
@@ -252,6 +252,9 @@ func (c *NodeController) watchResources(ctx context.Context, opt metav1.ListOpti
 		return err
 	}
 
+	// TODO: should record last resourceVersion for next round
+	opt.AllowWatchBookmarks = true
+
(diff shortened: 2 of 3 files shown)


kustomize/rbac/role.yaml
```diff
@@ -4,6 +4,18 @@ kind: ClusterRole
 metadata:
   name: kwok-controller
 rules:
+- apiGroups:
```
wzshiming, line 7: Add into 
https://github.com/kubernetes-sigs/kwok/blob/3df63f73098694fd0bf449353bce2669a0d7da01/pkg/apis/v1alpha1/doc.go#L21-L25

and run `./hack/update-controllergen.sh`

pkg/kwok/controllers/node_controller.go
```diff
@@ -252,6 +252,9 @@ func (c *NodeController) watchResources(ctx context.Context, opt metav1.ListOpti
 		return err
 	}
 
+	// TODO: should record last resourceVersion for next round
+	opt.AllowWatchBookmarks = true
```
wzshiming, line 256:
``` go
type resourceVersionGetter interface {
  GetResourceVersion() string
}
```

I'm not sure if adding this to the loop will achieve this effect?

``` go
opt.ResourceVersion = event.Object.(resourceVersionGetter).GetResourceVersion()
```

kustomize/rbac/role.yaml
fuweid: Updated

pkg/kwok/controllers/node_controller.go
fuweid: Basically, yes. Let me handle it in the follow-up, since I am trying to use existing informer pattern.

PR #870, opened 2023-12-15
chore: add unit test for utils/queue package

caozhuozi:
#### What type of PR is this?
/kind cleanup

Fixes #

Hi reviewers, 

Kwok is a truly awesome project! Thanks a lot for the contribution!

I hope I can join in and make it better (by just starting with some simple tasks :). This is my first attempt at contributing code to Kwok.

(I'm not sure such PR belongs to /kind cleanup. Please let me know if it is marked improperly.)

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/utils/queue/delaying_queue_test.go
+++ b/pkg/utils/queue/delaying_queue_test.go
@@ -1,0 +1,113 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package queue
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	fakeclock "k8s.io/utils/clock/testing"
+
+	"sigs.k8s.io/kwok/pkg/utils/wait"
+)
+
+func TestSimpleAddAfter(t *testing.T) {
+	fakeClock := fakeclock.NewFakeClock(time.Now())
+	dq := NewDelayingQueue[string](fakeClock)
+
+	first := "foo"
+	dq.AddAfter(first, 50*time.Millisecond)
+
+	// Simulate clock time passed
+	fakeClock.Step(10 * time.Millisecond)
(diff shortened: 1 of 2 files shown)


caozhuozi:
> /ok-to-test

Thanks, Shiming. Please feel free to give comments on my code if it is not good enough. I also noticed the regression test failed twice. Please let me know if it is caused by my code. Thanks!

pkg/utils/queue/delaying_queue_test.go
```diff
@@ -0,0 +1,113 @@
...
+	if !ok || x != first {
+		t.Fatalf("expected %s, got %s", first, x)
+	}
+}
+
+// checkAdded checks whether the delay queue has the expected number of elements.
+// The check action is supposed to be instantaneous.
+// However, in order to reserve time for the internal synchronization of the delay queue itself,
+// we perform the check in a polling manner and make it subject to a very short timeout(0.1s), the reserved time.
+func checkAdded[T comparable](q DelayingQueue[T], len int) error {
```
wzshiming, line 106: Possibly renamed checkLength to be more consistent with this functionality

pkg/utils/queue/delaying_queue_test.go
```diff
@@ -0,0 +1,113 @@
...
+func TestSimpleAddAfter(t *testing.T) {
+	fakeClock := fakeclock.NewFakeClock(time.Now())
+	dq := NewDelayingQueue[string](fakeClock)
+
+	first := "foo"
+	dq.AddAfter(first, 50*time.Millisecond)
+
+	// Simulate clock time passed
+	fakeClock.Step(10 * time.Millisecond)
+	if dq.Len() != 0 {
```
wzshiming, line 38: I think the length here may never increase because there is no real sleep time.

If possible, I would like to add a check to see if the length increases given sleep time.

wzshiming:
> Thanks, Shiming. Please feel free to give comments on my code if it is not good enough. I also noticed the regression test failed twice. Please let me know if it is caused by my code. Thanks!

Don't worry, I don't think it's the PR that's causing the flaky test, it's the lack of performance of the nested virtualization in the CI. It's not common for it to fail this many times, though.

wzshiming (commented):
Overall very good test

pkg/utils/queue/delaying_queue_test.go
caozhuozi: Sure.

pkg/utils/queue/delaying_queue_test.go
caozhuozi: Hi Shiming, I just added a  dedicated `checkIncreased` function for checking whether a queue is increased but with a sleep time given.

wzshiming:
/approve
/lgtm

Thank you for your contribution.

PR #877, opened 2023-12-19
Fix extended key usage for kwokctl

### LINKED ISSUE #869: unsuitable certificate purpose
### How to use it?

- [ ] kwok
- [X] kwokctl --runtime=docker (default runtime)
- [ ] kwokctl --runtime=binary
- [ ] kwokctl --runtime=kind

### What happened?

I found that curl cannot visit kwok cluster even if i supplied --cacert.

```
curl --cacert ~/.kwok/clusters/kwok/pki/ca.crt http://127.0.0.1:<port>
curl: (60) SSL certificate problem: unsuitable certificate purpose
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

it works fine in kubeadm clusters like kind or production clusters.

```
curl --cacert ~/.kwok/clusters/kwok/pki/ca.crt http://127.0.0.1:<port>
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "forbidden: User \"system:anonymous\" cannot get path \"/\"",
  "reason": "Forbidden",
  "details": {},
  "code": 403
}
```

just kwokctl create cluster, curl it (with openssl) or use rustls in rust, and you get it 

i think it's extended key usage that causing this problem.

which is related to this line(or file): 
https://github.com/wzshiming/kwok/blob/df21672368ff9046c00c98d9e44bb225de7ab903/pkg/kwokctl/pki/pki.go#L109C30-L109C41

key usage and extended key usage might need to be specified according to https://kubernetes.io/zh-cn/docs/setup/best-practices/certificates/

### Kwok version

<details>

```console
$ kwok --version
kwok version v0.4.0 go1.21.4 (darwin/amd64)

$ kwokctl --version
kwokctl version v0.4.0 go1.21.4 (darwin/amd64)
```

</details>

### OS version

<details>
```console
# On Darwin:
$ uname -a
Darwin ChaodeMacBook-Pro 23.1.0 Darwin Kernel Version 23.1.0: Mon Oct  9 21:27:27 PDT 2023; root:xnu-10002.41.9~6/RELEASE_X86_64 x86_64 i386 Darwin
```
</details>

jeffguorg:
#### What type of PR is this?

/kind bug

kwok apiserver cannot be accessed by other https clients for `invalid certificate puporse`

Fixes #869 

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kwokctl/pki/pki.go
+++ b/pkg/kwokctl/pki/pki.go
@@ -104,9 +104,15 @@ func GenerateSignCert(cn string, caCert *x509.Certificate, caKey crypto.Signer, 
 	}
 
 	certConfig := CertConfig{
-		CommonName:         cn,
-		Organization:       organizations,
-		Usages:             []x509.ExtKeyUsage{x509.ExtKeyUsageAny},
+		CommonName:   cn,
+		Organization: organizations,
+		Usages: []x509.ExtKeyUsage{
+			// TODO: replace any purpose with explicit EKU after each component get a separate certificate in kubernetes-sigs/kwok #878
+			x509.ExtKeyUsageAny,
+
+			x509.ExtKeyUsageServerAuth,
+			x509.ExtKeyUsageClientAuth,
+		},
 		AltNames:           alt,
 		PublicKeyAlgorithm: x509.RSA,
 		NotAfter:           notAfter,


jeffguorg:
i just realized the fix is not that simple. re-investigating how kwokctl creates certificates

/hold

pkg/kwokctl/pki/pki.go
```diff
@@ -106,7 +106,7 @@ func GenerateSignCert(cn string, caCert *x509.Certificate, caKey crypto.Signer,
 	certConfig := CertConfig{
 		CommonName:         cn,
 		Organization:       organizations,
-		Usages:             []x509.ExtKeyUsage{x509.ExtKeyUsageAny},
+		Usages:             []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
```
wzshiming, line 109:
```suggestion
		Usages:             []x509.ExtKeyUsage{
			x509.ExtKeyUsageAny,
			x509.ExtKeyUsageServerAuth,
		},
```

This should work.

but I'm not sure why ExtKeyUsageAny doesn't include ExtKeyUsageServerAuth, I've submitted an issue for this.
https://github.com/openssl/openssl/issues/23074

jeffguorg: I find that explicitly specify ExtKeyUsageServerAuth and ExtKeyUsageClientAuth should work.

i have updated the pr

jeffguorg:
i learned that kwokctl create one certificate for both apiserver and client usage. so i updated usages from server auth only to both server auth and client auth, including ext key usage from both [client](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/kubeconfig/kubeconfig.go#L215) and [apiserver](https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/certs/certlist.go#L275C53-L275C53).

/unhold

wzshiming:
I'm a little concerned that removing `x509.ExtKeyUsageAny` might cause an error, so it shouldn't be removed for now.

I think it will follow kubeadm's process of requesting separate certificates for each component, which I will do when I have time or when someone wants.

neolit123 (commented):
kubeadm only uses "any" for tests and chain verification:
https://github.com/kubernetes/kubernetes/blob/b5ba7bc4f5f49760c821cae2f152a8000922e72e/cmd/kubeadm/app/util/pkiutil/pki_helpers.go#L715
https://github.com/kubernetes/kubernetes/blob/b5ba7bc4f5f49760c821cae2f152a8000922e72e/cmd/kubeadm/app/util/certs/util.go#L154

"any" should work, so there could be a bug somwhere in the chain.
yet, being explicit about the key usages seems better.

wzshiming:
I agree with you, strange "any" works between Golang's PKI, but it seems OpenSSL doesn't respect it. whatever, change it to close to kubeadm behaviour in the future, I created an issue to track it #878

jeffguorg:
> whatever, change it to close to kubeadm behaviour in the future

@wzshiming sure. i added an any purpose alongside server & client auth to keep the previous behavior. does it look good now?

wzshiming:
/approve
/lgtm

Thank you for your contribution!

PR #879, opened 2023-12-20
fix(kwokctl): modify misunderstanding description in config subcommand

caozhuozi:
#### What type of PR is this?

/kind documentation
/kind cleanup

the description of the `config view` and `config tidy` subcommands may lead to misunderstanding. 
Specifically,  the description of the 2 commands says "Display/Tidy the default config file **with –config**", from which somebody might think  "`--config`" is required to show/tidy the default configuration. However, "`--config`" is used to set the config file path in `kwok`. 

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kwokctl/cmd/config/tidy/tidy.go
+++ b/pkg/kwokctl/cmd/config/tidy/tidy.go
@@ -33,7 +33,7 @@ func NewCommand(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
 		Args:  cobra.NoArgs,
 		Use:   "tidy",
-		Short: "Tidy the default config file with --config",
+		Short: "Tidy the default config file",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runE(cmd.Context())
 		},
--- a/pkg/kwokctl/cmd/config/tidy/tidy.go
+++ b/pkg/kwokctl/cmd/config/tidy/tidy.go
@@ -33,7 +33,7 @@ func NewCommand(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
 		Args:  cobra.NoArgs,
 		Use:   "tidy",
-		Short: "Tidy the default config file",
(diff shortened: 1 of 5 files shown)


pkg/kwokctl/cmd/config/tidy/tidy.go
```diff
@@ -33,7 +33,7 @@ func NewCommand(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
 		Args:  cobra.NoArgs,
 		Use:   "tidy",
-		Short: "Tidy the default config file with --config",
+		Short: "Tidy the default config file",
```
wzshiming, line 36: Here the `--config` actually merges the default configuration with the configuration it specifies into the default configuration

pkg/kwokctl/cmd/config/view/view.go
```diff
@@ -34,7 +34,7 @@ func NewCommand(ctx context.Context) *cobra.Command {
 	cmd := &cobra.Command{
 		Args:  cobra.NoArgs,
 		Use:   "view",
-		Short: "Display the default config file with --config",
+		Short: "Display the default config file",
```
wzshiming, line 37: This is similar to the above, without changing the default configuration, just output it

wzshiming:
Also, need to use `./hack/update-cmd-docs.sh` to regenerate the docs.

pkg/kwokctl/cmd/config/tidy/tidy.go
caozhuozi: Good point! Thanks, Shiming.

caozhuozi:
> Also, need to use `./hack/update-cmd-docs.sh` to regenerate the docs.

Oh, Thanks for pointing out! I originally thought the docs are auto-generated.

caozhuozi:
Hi, Shiming. I just pushed a new commit. I'm not sure if using periods in `Short` breaks the convention.

Issue #816, opened 2023-10-16
[kwok] controller doesn't retry when the stage play runs into transient error


fuweid:
### How to use it?

- [X] kwok
- [ ] kwokctl --runtime=docker (default runtime)
- [ ] kwokctl --runtime=binary
- [ ] kwokctl --runtime=nerdctl
- [ ] kwokctl --runtime=kind

### What happened?

```
{"time":"2023-10-16T02:35:26.331648934Z","level":"ERROR","source":{"function":"sigs.k8s.io/kwok/pkg/kwok/controllers.(*PodController).playStage","file":"/workspace/pkg/kwok/controllers/pod_controller.go","line":346},"msg":"Failed to patch node","id":"kwok-virtualnode-42_383d16ce-8002-4b33-8774-df7428be80b6","pod":"argo/stress-883-sleep-3884558001","node":"kwok-virtualnode-42","stage":"pod-init-container-completed","err":"Patch \"https://10.0.0.1:443/api/v1/namespaces/argo/pods/stress-883-sleep-3884558001/status\": read tcp 10.244.1.37:36720->10.0.0.1:443: read: connection reset by peer"}
```

When the controller has transient error, it won't retry the pod stage. So that the pod will be stuck at the current stage forever.

> NOTE: Typo s/node/pod at https://github.com/kubernetes-sigs/kwok/blob/db6cea86d35b03c3694bd332c7f92f4fe1a05f24/pkg/kwok/controllers/pod_controller.go#L346

### What did you expect to happen?

The controller should retry the next stage play until it's successful

### How can we reproduce it (as minimally and precisely as possible)?

1. apply normal pod and the stage will move the stage into completed after 5s.
2. before the next stage, restart the apiserver and let the controller hits the error
3. 

### Kwok version

v0.4.0

### OS version

 Linux ubuntu 5.15.0-1049-azure

caozhuozi:
Hi, @wzshiming. I think the issue might be solved if we reput the resource into `preprocessChan` if an error occurs?

(The `delay` should be ignored for a retry.)

https://github.com/kubernetes-sigs/kwok/blob/db6cea86d35b03c3694bd332c7f92f4fe1a05f24/pkg/kwok/controllers/pod_controller.go#L345-L347

wzshiming:
> Hi, @wzshiming. I think the issue might be solved if we reput the resource into `preprocessChan` if an error occurs?
> 
> (The `delay` should be ignored for a retry.)
> 
[...]

PR #904, opened 2024-01-04
[KWOK] add retry feature when fails to play stages

caozhuozi:
#### What type of PR is this?

/kind feature

The PR tries to fix https://github.com/kubernetes-sigs/kwok/issues/816

Fixes https://github.com/kubernetes-sigs/kwok/issues/816

**The feature is still in progress**. 

**In the first stage**, I am going to provide the related data structure for review.
After it is revised and approved by you, I will then start to change the controller logic.

The data structure I'm going to use to implement the retry feature  is called `retry queue`, which combines a delaying queue and a `backoffSet` to help track the backoff delay time of all the items.

Please see the code for details. Feel free to give comments.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/utils/backoff/backoff_set.go
+++ b/pkg/utils/backoff/backoff_set.go
@@ -1,0 +1,226 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package backoff
+
+import (
+	"sync"
+	"time"
+
+	"sigs.k8s.io/kwok/pkg/utils/wait"
+)
+
+const (
+	DefaultInitialInterval time.Duration = 10 * time.Second
+	DefaultMaxInterval     time.Duration = 5 * time.Minute
+	DefaultFactor          float64       = 2.0
+	DefaultJitter          float64       = 0.1
+)
+
+// Set is used for tracking multiple items that shares a same backoff setting.
+// Set is inspired by k8s.io/client-go/util/flowcontrol/backoff.go.
+// In addition to re implementing it with in a generic style, we also simplify
+// the functionality as much as possible according to our own needs.
+type Set[T comparable] struct {
+	sync.RWMutex
+	// the init backoff period of a new added item is initialInterval plus jitter
+	initialInterval time.Duration
+	// Factor is the multiplying factor for each revision on the backoff duration.
+	factor float64
+	// The (revised) backoff duration is capped at cap.
+	cap time.Duration
+	// Jitter adds a randomized amount of time to the revised duration.
+	// The amount is chosen uniformly at random from the interval between
+	// zero and `jitter*current revised duration`.
+	// No jitter will be added if jitter is zero.
+	jitter float64
+	// entries holds the current backoff duration of all the items.
+	entries map[T]time.Duration
+}
+
+// SetOptions configures the Set
+type SetOptions struct {
+	initialInterval time.Duration
+	factor          float64
+	maxInterval     time.Duration
+	jitter          float64
+}
+
+// OptionFunc configures a SetOption
+type OptionFunc func(*SetOptions)
+
+// WithInitialInterval configures the initialInterval for the backoff.
+// DefaultInitialInterval will be used instead if given a negative augment.
+func WithInitialInterval(interval time.Duration) OptionFunc {
+	if interval < 0 {
+		interval = DefaultInitialInterval
+	}
+	return func(options *SetOptions) {
+		options.initialInterval = interval
+	}
+}
+
+// WithMaxInterval limits the max period for the backoff.
+// DefaultMaxInterval will be used instead if given a non-positive augment.
+func WithMaxInterval(maxInterval time.Duration) OptionFunc {
+	if maxInterval <= 0 {
+		maxInterval = DefaultMaxInterval
+	}
+	return func(opt *SetOptions) {
+		opt.maxInterval = maxInterval
+	}
+}
+
+// WithFactor configures the factor for the backoff.
+// DefaultFactor will be used instead if given a non-positive augment.
+func WithFactor(factor float64) OptionFunc {
+	if factor <= 0.0 {
+		factor = DefaultFactor
+	}
+	return func(opt *SetOptions) {
+		opt.factor = factor
+	}
+}
+
+// WithJitter configures the jitter for the backoff.
+// DefaultJitter will be used instead if given a negative augment.
+func WithJitter(jitter float64) OptionFunc {
+	if jitter < 0.0 {
+		jitter = DefaultJitter
+	}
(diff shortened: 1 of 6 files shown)


pkg/utils/queue/delaying_queue_test.go
```diff
@@ -104,9 +104,9 @@ func TestAddTwo(t *testing.T) {
 
 // checkLength checks whether the delay queue has the expected number of elements.
 // The check action is supposed to be instantaneous.
-// However, in order to reserve time for the internal synchronization of the delay queue itself,
+// However, in order to reserve time for the internal synchronization of the (delay) queue itself,
 // we perform the check in a polling manner and make it subject to a very short timeout(0.1s), the reserved time.
-func checkLength[T comparable](q DelayingQueue[T], len int) error {
+func checkLength[T comparable](q Queue[T], len int) error {
```
wzshiming, line 109:
```suggestion
func checkLength(q length, len int) error {
```

Define a generic interface for Len()
``` golang
type length interface {
	Len() int
}
```

pkg/utils/queue/retry_queue.go
```diff
@@ -0,0 +1,118 @@
...
+// a retry queue is re-defined.
+// The caller calls Get to peek an item to retry, but the item should
+// not be removed from the underlying state unless the
+// caller explicitly calls the Remove method. The caller MUST also
+// explicitly re-Add the item to the queue if a retry fails or Remove it
+// if it succeeds.
+type RetryQueue[T comparable] interface {
+	Queue[T]
+	// Remove excludes an item from being tracked.
+	Remove(T)
```
wzshiming, line 39:
```suggestion
```

Overloaded the `Cancel` instead add the `Remove`

pkg/utils/queue/retry_queue.go
```diff
@@ -0,0 +1,118 @@
...
+
+// RetryQueue abstracts the behaviour of a queue for retry purpose.
+// Compared to a general Queue type, the meaning of the methods of
+// a retry queue is re-defined.
+// The caller calls Get to peek an item to retry, but the item should
+// not be removed from the underlying state unless the
+// caller explicitly calls the Remove method. The caller MUST also
+// explicitly re-Add the item to the queue if a retry fails or Remove it
+// if it succeeds.
+type RetryQueue[T comparable] interface {
```
wzshiming, line 36: It doesn't feel like a queue anymore. Wouldn't it be more appropriate to put it in "sigs.k8s.io/kwok/pkg/utils/wait"?

pkg/utils/backoff/backoff_set.go
```diff
@@ -0,0 +1,226 @@
...
+limitations under the License.
+*/
+
+package backoff
+
+import (
+	"sync"
+	"time"
+
+	"sigs.k8s.io/kwok/pkg/utils/wait"
```
wzshiming, line 23: It feels like dependence is reversed

caozhuozi: @wzshiming  Yes. This will cause an import cycle if we put `RetryQueue` in `wait` package. Any suggestion about the layout? Or if it's not feasible, I can also refactor the code based on your suggestions.

the current implementation is  based on your comments:
https://github.com/kubernetes-sigs/kwok/issues/816#issuecomment-1869461211

wzshiming: https://github.com/kubernetes-sigs/kwok/pull/902/files#diff-b0dafdbd30f76f38d1f10f9abfdbb4adda9d8eaa2d674ef18134ac7d50c0bdf7R113-R139

In this PR, I have tried to replace Normal Queue and Failed Queue with Priority Queue. It works pretty well with the proportional processing of events.


then the number of failures of the operation is recorded instead of the number of seconds of the last sleep, and the number of seconds of the next sleep is calculated based on the number of failures each time.

caozhuozi: Thanks @wzshiming.  Will dig into #902 and digest what you said here.

caozhuozi: Hi, @wzshiming. Do you mean once a new event of a resource **type**(not a **individual** resource) comes in, the **entire** failed queue should be canceled? So there is no need to maintain a failed queue, right?

wzshiming: Yes and the weight of the failure queue should be the lowest

wzshiming: In the case of a single queue, normal will be directly blocked once a large batch of updates fail. Although separating the normal and failure queues alleviates this issue to some extent, it still results in the blocking of the normal. To address this, we enforce a normal queue using the weighted queue.

caozhuozi: > The overall design is to prioritize new events and **failures are only processed when there are no new events**.

@wzshiming  Maybe I misunderstood what you meant here.

As per the implementation of the weighted queue, If we put the failed tasks into the queue with the lowest weight (`1`), **they still have a chance to be processed**  if new events are coming **but just with a relatively low proportion**. This is somewhat contradictory to what you said (quoted) 

Could you please further clarify it? Thanks!

caozhuozi: Hi, @wzshiming 

The following are my understanding for the retry feature:

1. replace the delay queue with weight delay queue
2. for a normal(fresh) stage job, push it in the weighted delay queue with weight `0` (highest priority)
3. when *`playStage()`* fails in *`playStageWorker`*, we repush it immediately to the weight delay queue with weight `1` as well as a **backoff delay**. The backoff period should be calculated  based on the **failed times** (according to your new comments). Thus we also need a dediated map to track the failed times of each job. My question here:
   
    - **do we need a dedicated struct for this map? (since the associated  backoff settings(cap, factor, jitter) also need to be recorded.)**
    - If above if true, **where should I put the instance of this struct?** Should I put it directly under the Controller (`PodController`, `NodeController`) as the `delayQueueMapping` one?
    
Please correct me if there is any misunderstanding. Thanks again for your great patience! ;-)

wzshiming: Thank you, in my opinion, maybe keep track of the number of failures. then calculate the next time when failed, such as `jitter(min(100ms * (2 ** failed_number), maximal_backoff), jitter_factor)`.

caozhuozi: I have an idea that:

 Using Job  struct  *`resourceStageJob`* itself to record the failed count instead of using an additional dedicated map to track the failed counts.

I think there may several drawbacks if using an additional  map:

Since we already have an associated map struct `delayQueueMapping,` maintaining too many scattered map structs around the queue is not a good idea. For example, when we cancel a job, we also need to remove the associated item from that map. https://github.com/kubernetes-sigs/kwok/blob/main/pkg/kwok/controllers/pod_controller.go#L252

Also, if we add a failed job to the queue, we need to add them both in `delayQueueMapping` and the failed count map.


In terms of implementation, I want to add  a field `RetryCount`  to *`resourceStageJob`* :

```go
type resourceStageJob[T any] struct {
	Resource T
	Stage    *LifecycleStage
	Key      string
	
        RetryCount int
}
```

When a job is pushed to the queue for the first time, the `RetryCount` should be zero. Once  *`playStage()`* fails in *`playStageWorker` ,* we incease the failed count and repush it to the queue with weight `1` and a delay time, which is calculated based on the increased retry times.


What do you think?

wzshiming: Good idea.

Note that this requires using a pointer and initializing it, otherwise it won't match the original value
``` golang
RetryCount *int
```

wzshiming:
/close

Fixed by #911

PR #911, opened 2024-01-11
[KWOK] retry when applying Stages fails

caozhuozi:
#### What type of PR is this?
/kind feature

Fixes https://github.com/kubernetes-sigs/kwok/issues/816

Since we have a new design based on the weight delaying queue(https://github.com/kubernetes-sigs/kwok/pull/902), the old implementation https://github.com/kubernetes-sigs/kwok/pull/904  was rejected.   Will close the old one and start here. 

Currently, I **only modified the node controller** for a pre-review. Please feel free to give comments!

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/kwok/controllers/node_controller.go
+++ b/pkg/kwok/controllers/node_controller.go
@@ -384,7 +384,7 @@ func (c *NodeController) preprocess(ctx context.Context, node *corev1.Node) erro
 		Resource:   node,
 		Stage:      stage,
 		Key:        key,
-		RetryCount: new(int),
+		RetryCount: new(uint64),
 	}
 	// we add a normal(fresh) stage job with weight 0,
 	// resulting in that it will always be processed with high priority compared to those retry ones
@@ -410,15 +410,15 @@ func (c *NodeController) playStageWorker(ctx context.Context) {
 			)
 		}
 		if needRetry {
-			*node.RetryCount++
+			retryCount := atomic.AddUint64(node.RetryCount, 1) - 1
 			logger.Info("retrying for failed job",
 				"node", node.Key,
 				"stage", node.Stage.Name(),
-				"retry", *node.RetryCount,
+				"retry", retryCount,
 			)
 			// for failed jobs, we re-push them into the queue with a lower weight
 			// and a backoff period to avoid blocking normal tasks
-			retryDelay := backoffDelayByStep(*node.RetryCount, c.backoff)
+			retryDelay := backoffDelayByStep(retryCount, c.backoff)
 			c.addStageJob(node, retryDelay, 1)
 		}
 	}
@@ -467,8 +467,8 @@ func (c *NodeController) playStage(ctx context.Context, node *corev1.Node, stage
 			return shouldRetry(err), fmt.Errorf("failed to compute the status patch of node %s: %w", node.Name, err)
 		}
 		if patch == nil {
-			logger.Debug("Skip modify status",
-				"reason", "do not need to modify status",
+			logger.Debug("Skip node",
+				"reason", "do not need to modify",
 			)
 		} else {
 			result, err = c.patchResource(ctx, node, patch)
--- a/pkg/kwok/controllers/node_controller.go
+++ b/pkg/kwok/controllers/node_controller.go
@@ -23,9 +23,9 @@ import (
 	"fmt"
 	"net"
 	"sync/atomic"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
@@ -43,6 +43,7 @@ import (
 	"sigs.k8s.io/kwok/pkg/utils/informer"
 	"sigs.k8s.io/kwok/pkg/utils/maps"
 	"sigs.k8s.io/kwok/pkg/utils/queue"
+	"sigs.k8s.io/kwok/pkg/utils/wait"
 )
 
 var (
@@ -98,8 +99,9 @@ type NodeController struct {
 	preprocessChan                        chan *corev1.Node
 	playStageParallelism                  uint
 	lifecycle                             resources.Getter[Lifecycle]
-	delayQueue                            queue.DelayingQueue[resourceStageJob[*corev1.Node]]
+	delayQueue                            queue.WeightDelayingQueue[resourceStageJob[*corev1.Node]]
 	delayQueueMapping                     maps.SyncMap[string, resourceStageJob[*corev1.Node]]
+	backoff                               wait.Backoff
 	recorder                              record.EventRecorder
 	readOnlyFunc                          func(nodeName string) bool
 	enableMetrics                         bool
@@ -159,7 +161,8 @@ func NewNodeController(conf NodeControllerConfig) (*NodeController, error) {
 		nodeIP:                                conf.NodeIP,
 		nodeName:                              conf.NodeName,
 		nodePort:                              conf.NodePort,
-		delayQueue:                            queue.NewDelayingQueue[resourceStageJob[*corev1.Node]](conf.Clock),
+		delayQueue:                            queue.NewWeightDelayingQueue[resourceStageJob[*corev1.Node]](conf.Clock),
+		backoff:                               defaultBackoff(),
 		lifecycle:                             conf.Lifecycle,
 		playStageParallelism:                  conf.PlayStageParallelism,
 		preprocessChan:                        make(chan *corev1.Node),
@@ -264,7 +267,7 @@ loop:
 	logger.Info("Stop watch nodes")
 }
 
-// finalizersModify modify finalizers of node
+// finalizersModify modifies the finalizers of a node
 func (c *NodeController) finalizersModify(ctx context.Context, node *corev1.Node, finalizers *internalversion.StageFinalizers) (*corev1.Node, error) {
 	ops := finalizersModify(node.Finalizers, finalizers)
 	if len(ops) == 0 {
@@ -280,14 +283,8 @@ func (c *NodeController) finalizersModify(ctx context.Context, node *corev1.Node
 		"node", node.Name,
 	)
 
-	result, err := c.typedClient.CoreV1().Nodes().Patch(ctx, node.Name, types.JSONPatchType, data, metav1.PatchOptions{})
+	result, err := c.patchResource(ctx, node, types.JSONPatchType, data)
 	if err != nil {
-		if apierrors.IsNotFound(err) {
-			logger.Warn("Patch node finalizers",
-				"err", err,
-			)
-			return nil, nil
-		}
 		return nil, err
 	}
 	logger.Info("Patch node finalizers")
@@ -303,12 +300,6 @@ func (c *NodeController) deleteResource(ctx context.Context, node *corev1.Node) 
 
 	err := c.typedClient.CoreV1().Nodes().Delete(ctx, node.Name, deleteOpt)
 	if err != nil {
-		if apierrors.IsNotFound(err) {
-			logger.Warn("Delete node",
-				"err", err,
-			)
-			return nil
-		}
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/kwok/controllers/node_controller.go
```diff
@@ -615,3 +632,21 @@ func (c *NodeController) funcNodeName() string {
...
+
+	delay := float64(c.backoff.Duration)
+
+	for i := 0; i < failedTimes; i++ {
+		delay *= c.backoff.Factor
+		if delay >= float64(c.backoff.Cap) {
+			delay = float64(c.backoff.Cap)
+			break
+		}
+	}
```
wzshiming, line 650:
```suggestion
	delay := math.Min(float64(c.backoff.Duration) * math.Pow(c.backoff.Factor, float64(failedTimes)), c.backoff.Cap)
```

pkg/kwok/controllers/node_controller.go
```diff
@@ -615,3 +632,21 @@ func (c *NodeController) funcNodeName() string {
 func (c *NodeController) funcNodePort() int {
 	return c.nodePort
 }
+
+// backoffDelay calculates the backoff delay of a Stage job based on its failed times
+func (c *NodeController) backoffDelay(failedTimes int) time.Duration {
```
wzshiming, line 637: It feels like this function is moving to utils and will be shared for node, pod and stage controllers.

pkg/kwok/controllers/node_controller.go
```diff
@@ -403,15 +412,26 @@ func (c *NodeController) preprocess(ctx context.Context, node *corev1.Node) erro
...
 // playStageWorker receives the resource from the playStageChan and play the stage
 func (c *NodeController) playStageWorker(ctx context.Context) {
+	logger := log.FromContext(ctx)
+
 	for ctx.Err() == nil {
 		node := c.delayQueue.GetOrWait()
 		c.delayQueueMapping.Delete(node.Key)
-		c.playStage(ctx, node.Resource, node.Stage)
+		err := c.playStage(ctx, node.Resource, node.Stage)
+		if err != nil {
```
wzshiming, line 421: Certain errors should not be retried, such as notfound, conflict, and the patch itself being misformatted

pkg/kwok/controllers/node_controller.go
```diff
@@ -389,10 +394,14 @@ func (c *NodeController) preprocess(ctx context.Context, node *corev1.Node) erro
...
-		Resource: node,
-		Stage:    stage,
-		Key:      key,
+		Resource:   node,
+		Stage:      stage,
+		Key:        key,
+		RetryCount: &retryTimes,
 	}
 
 	c.delayQueue.AddAfter(item, delay)
```
wzshiming, line 407: Maybe we should wrap the following into methods and try to de-duplicate multiple events on the same object

``` golang
c.delayQueue.AddWeightAfter(...)
c.delayQueueMapping.Store(...)
```

pkg/kwok/controllers/node_controller.go
caozhuozi: I use a raw for loop here since  `failedTimes`  may grow very large and cause `math.Pow` to overflow?

wzshiming: `math.Pow` too large returns +Inf, which can also be used for `math.Min`, 

``` golang
math.Pow(2, 10000) # +Inf
math.Min(math.Pow(2, 10000), 100) # 100
```

pkg/kwok/controllers/node_controller.go
caozhuozi: I also have an idea that if we only retry when there are network issues, 
Can we just enable **resync** in informer such that we can re-process events (constantly) after network issues are gone? If we, we don't need retry feature at all.

wzshiming: > I also have an idea that if we only retry when there are network issues,
Could we enable resync in informer such that we can re-process events (constantly) after network recovers? If so, we may don't need a retry feature.


This way is not possible, for two reasons:
- For save memory, [does not store data in memory](https://github.com/kubernetes-sigs/kwok/blob/f81396847566ec7246c23331e5235b6be8011cc2/pkg/utils/informer/informer.go#L163) in many cases
- It is not suitable for large amounts of data

caozhuozi:
Hi, @wzshiming.
I added a `shoudRetry` function to determine whether an error needs to be retried.

Currently, we only check those network errors and bypass other error types.

wzshiming:
I think like We can use `errors.Join` and `fmt.Errorf`

caozhuozi:
Hi, @wzshiming. Do you mean to join the specific error with the newly defined error?
I have two concerns here:
1. I'm not sure if it is a **Go idiomatic** way
2. the other concern is that where should I put the `join` statement? There are 2 choices currently:
    * inside `computePatch` method, like:

      ```go
        func (c *NodeController) computePatch(node *corev1.Node, tpl string) ([]byte, error) {
        patch, err := c.renderer.ToJSON(tpl, node)
        if err != nil {
        return errors.Join(ComputePatchError, err)  // the original `err` will be lost
        }
        // omitted
        }
      ```
    *  at where `computePath` method is called, like

       ```go
       patch, err = c.computePatch(node, next.StatusTemplate)
       if err != nil {
           return errors.Join(ComputePatchError, err)
       }
       ```
       
   which one do you prefer?

wzshiming:
I think you can finish this and verify the various corner cases locally

caozhuozi:
I made several changes in this commit:
1. define a new error type `ErrComputePatch`
2. remove those resource `isNotFound` checks as they will be handled in `shouldRetry`, which is the **only entry** for determining errors.
3. move the `patch` operation out of `finalizersModify` and rename the method to `computeFinalizerPatch`. If any error occurs in `computeFinalizerPatch`, we should also mark the error as `ErrComputePatch`.
4. rename the `computePatch` to `computeStatusPatch`.
5. generalize `patchResource` method such that it can be used by patching status and patching finalizers.

caozhuozi:
since we only need to check the network error, I just removed the defined `ErrComputePatch` in the latest commit.

pkg/kwok/controllers/node_controller.go
```diff
@@ -615,3 +620,9 @@ func (c *NodeController) funcNodeName() string {
 func (c *NodeController) funcNodePort() int {
 	return c.nodePort
 }
+
+// addStageJob adds a stage to be applied into the underlying weight delay queue and the associated helper map
+func (c *NodeController) addStageJob(job resourceStageJob[*corev1.Node], delay time.Duration, weight int) {
+	c.delayQueue.AddWeightAfter(job, weight, delay)
+	c.delayQueueMapping.Store(job.Key, job)
```
wzshiming, line 627:
```suggestion
	old, loaded := c.delayQueueMapping.Swap(job.Key, job)
	if loaded {
		c.delayQueue.Cancel(old)
	}
	c.delayQueue.AddWeightAfter(job, weight, delay)
```

pkg/kwok/controllers/node_controller.go
```diff
@@ -168,6 +170,8 @@ func NewNodeController(conf NodeControllerConfig) (*NodeController, error) {
 		enableMetrics:                         conf.EnableMetrics,
 	}
 
+	c.backoff = wait.Backoff{Duration: 100 * time.Second, Factor: 2.0, Jitter: 0.1, Cap: 5 * time.Minute}
```
wzshiming, line 173: Move to utils.go, shared by Node, Pod and Stage

caozhuozi:
In this new commit,
* Get the  `finalizerModify` logic back
* check whether the key already exists in the map when adding a job
* Create a function in `utils` to obtain a default backoff setting shared by all kwok controllers for those retried jobs.
* rebase the code based on the recent changes.

If there are no problems with this version, I will start to change other controllers.

pkg/kwok/controllers/node_controller.go
```diff
@@ -98,8 +99,9 @@ type NodeController struct {
 	preprocessChan                        chan *corev1.Node
 	playStageParallelism                  uint
 	lifecycle                             resources.Getter[Lifecycle]
-	delayQueue                            queue.DelayingQueue[resourceStageJob[*corev1.Node]]
+	delayQueue                            queue.WeightDelayingQueue[resourceStageJob[*corev1.Node]]
```
wzshiming, line 102: We should add the same support for Pod and Stage

caozhuozi: Hi @wzshiming  Yes. I will change other controllers as well If there are no obvious problems with this version.

wzshiming:
Sorry for the delay, this PR may need to be rebase due to the conflict with #920, I will take time to test this PR this week.

caozhuozi:
> Sorry for the delay, this PR may need to be rebase due to the conflict with #920, I will take time to test this PR this week.

Oh, @wzshiming! Understood! No worry! I will rebase it. 😊

pkg/kwok/controllers/stage_controller.go
```diff
@@ -310,32 +315,28 @@ func (c *StageController) playStage(ctx context.Context, resource *unstructured.
...
-			return
+			return shouldRetry(err), fmt.Errorf("failed to compute the status patch of resource %s: %w", resource.GetName(), err)
 		}
 		if patch == nil {
 			logger.Debug("Skip resource",
 				"reason", "do not need to modify",
 			)
 		} else {
-			result, err = c.patchResource(ctx, resource, patch)
+			result, err = c.patchResource(ctx, resource, types.StrategicMergePatchType, patch, "status")
```
wzshiming, line 337:
```suggestion
			result, err = c.patchResource(ctx, resource, types.MergePatchType, patch, "status")
```

This is not the same as Pod and Node, using `types.StrategicMergePatchType` will cause patch failure in some cases, but I forget the specific reason.

pkg/kwok/controllers/error.go
```diff
@@ -0,0 +1,23 @@
+package controllers
+
+import (
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
+)
+
+// shouldRetry determines if a certain error needs to be retried
+func shouldRetry(err error) bool {
```
wzshiming, line 9: move to utils.go

pkg/kwok/controllers/stage_controller.go
```diff
@@ -264,32 +255,51 @@ func (c *StageController) preprocess(ctx context.Context, resource *unstructured
 		)
 	}
 
+	// We initialize a normal(fresh) job with a zero `RetryCount`.
+	// The counter may increase if the job fails to apply.
```
wzshiming, line 259: ditto

pkg/kwok/controllers/pod_controller.go
```diff
@@ -284,32 +275,50 @@ func (c *PodController) preprocess(ctx context.Context, pod *corev1.Pod) error {
...
-		c.playStage(ctx, pod.Resource, pod.Stage)
+		needRetry, err := c.playStage(ctx, pod.Resource, pod.Stage)
+		if err != nil {
+			logger.Error("failed to apply stage", err,
+				"node", pod.Resource.Name, "stage", pod.Stage.Name())
+		}
+		if needRetry {
+			*pod.RetryCount++
+			logger.Info("retrying for failed job",
+				"pod", pod.Resource.Name, "stage", pod.Stage.Name(), "retry", *pod.RetryCount)
```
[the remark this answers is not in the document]
caozhuozi: Also should be `"pod", pod.Key`, here?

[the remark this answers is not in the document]
wzshiming: Yea

pkg/kwok/controllers/stage_controller.go
```diff
@@ -265,31 +256,53 @@ func (c *StageController) preprocess(ctx context.Context, resource *unstructured
...
+		if needRetry {
+			*resource.RetryCount++
+			logger.Info("retrying for failed job",
+				"resource", resource.Key,
+				"stage", resource.Stage.Name(),
+				"retry", *resource.RetryCount,
+			)
+			// for failed jobs, we re-push them into the queue with a lower weight
+			// and a backoff period to avoid blocking normal tasks
+			retryDelay := backoffDelayByStep(*resource.RetryCount, c.backoff)
```
wzshiming, line 297: ditto

pkg/kwok/controllers/utils.go
```diff
@@ -112,4 +117,40 @@ type resourceStageJob[T any] struct {
...
+}
+
+// defaultBackoff provides a backoff setting for kwok controllers to apply failed jobs
+func defaultBackoff() wait.Backoff {
+	return wait.Backoff{Duration: 1 * time.Second, Factor: 2.0, Jitter: 0.2, Cap: 32 * time.Minute}
+}
+
+// backoffDelayByStep calculates the backoff delay period based on steps
+func backoffDelayByStep(steps int, c wait.Backoff) time.Duration {
+	if steps <= 0 {
```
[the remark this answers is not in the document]
caozhuozi: since `uint64` has no value less than zero after the newest change, I just removed the `if` clause.

pkg/kwok/controllers/pod_controller.go
```diff
@@ -285,31 +276,52 @@ func (c *PodController) preprocess(ctx context.Context, pod *corev1.Pod) error {
...
+			*pod.RetryCount++
+			logger.Info("retrying for failed job",
+				"pod", pod.Key,
+				"stage", pod.Stage.Name(),
+				"retry", *pod.RetryCount,
+			)
+			// for failed jobs, we re-push them into the queue with a lower weight
+			// and a backoff period to avoid blocking normal tasks
+			retryDelay := backoffDelayByStep(*pod.RetryCount, c.backoff)
+			c.addStageJob(pod, retryDelay, 1)
```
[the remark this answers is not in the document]
caozhuozi: Thanks!!

caozhuozi:
@wzshiming 

I made many mistakes and misunderstood a lot of things along the way (and I also learned a lot.) Thanks for the great patience and for taking the trouble to guide me during the PR.

PR #964, opened 2024-02-06
Support cors

sivchari:
#### What type of PR is this?

/kind feature

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
Add "kube-apiserver-cors-allowed-origin" flag to specify cors policy'
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #971, opened 2024-02-09
Support extra args

sivchari:
#### What type of PR is this?

related #964 

/kind feature

I support extra-args for kwokctl

Fixes #

```release-note
Support --extra-args
```

```docs

```


pkg/kwokctl/runtime/kind/cluster.go
```diff
@@ -378,11 +378,50 @@ func (c *Cluster) addKind(ctx context.Context, env *env) (err error) {
...
+			Name: splitedArgs[0],
+			ExtraArgs: []internalversion.ExtraArgs{
+				{
+					Key:   splitedArgs[1],
+					Value: splitedArgs[2],
+				},
+			},
+		})
+		componentNames[splitedArgs[0]] = i
+	}
```
wzshiming, line 405: I think we can move this part into main and add a convert from string to ComponentPatches, like
https://github.com/kubernetes-sigs/kwok/blob/b8917eccea1e0eb80af5074787573d86a6ca3231/pkg/kwokctl/cmd/create/cluster/cluster.go#L190

sivchari: I've done.

sivchari:
Do you know why some tests are failed ?

pkg/kwokctl/runtime/kind/cluster.go
```diff
@@ -379,10 +379,30 @@ func (c *Cluster) addKind(ctx context.Context, env *env) (err error) {
...
+			args := filterDuplicatedExtraArgs(kubeSchedulerComponentPatches.ExtraArgs, patch.ExtraArgs)
+			kubeSchedulerComponentPatches.ExtraArgs = args
+		case consts.ComponentKubeControllerManager:
+			args := filterDuplicatedExtraArgs(kubeControllerManagerComponentPatches.ExtraArgs, patch.ExtraArgs)
+			kubeControllerManagerComponentPatches.ExtraArgs = args
+		case consts.ComponentKwokController:
+			args := filterDuplicatedExtraArgs(kwokControllerComponentPatches.ExtraArgs, patch.ExtraArgs)
+			kwokControllerComponentPatches.ExtraArgs = args
+		}
+	}
```
wzshiming, line 405: This will change the expected behavior.

We may need to design features that override existing args.

WDYT?

pkg/apis/config/v1alpha1/kwokctl_configuration_types.go
```diff
@@ -454,6 +454,9 @@ type KwokctlConfigurationOptions struct {
 	// DisableQPSLimits specifies whether to disable QPS limits for components.
 	// +default=false
 	DisableQPSLimits *bool `json:"disableQPSLimits,omitempty"`
+
+	// ExtraArgs is used to pass extra arguments to the components.
+	ExtraArgs []string `json:"extraArgs,omitempty"`
```
wzshiming, line 459: I feel this as a configuration should be inappropriate maybe we should move it to

https://github.com/kubernetes-sigs/kwok/blob/b8917eccea1e0eb80af5074787573d86a6ca3231/pkg/kwokctl/cmd/create/cluster/cluster.go#L38-L45

pkg/kwokctl/cmd/create/cluster/cluster.go
```diff
@@ -145,6 +145,7 @@ func NewCommand(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringSliceVar(&flags.Options.EnableCRDs, "enable-crds", flags.Options.EnableCRDs, "List of CRDs to enable")
 	cmd.Flags().UintVar(&flags.Options.NodeLeaseDurationSeconds, "node-lease-duration-seconds", flags.Options.NodeLeaseDurationSeconds, "Duration of node lease in seconds")
 	cmd.Flags().Float64Var(&flags.Options.HeartbeatFactor, "heartbeat-factor", flags.Options.HeartbeatFactor, "Scale factor for all about heartbeat")
+	cmd.Flags().StringArrayVar(&flags.Options.ExtraArgs, "extra-args", flags.Options.ExtraArgs, "A set of key=value pairs that is passed to the runtime")
```
wzshiming, line 148: Should we have a detailed explanation of the `component=key=value` format?

pkg/kwokctl/components/utils.go
```diff
@@ -26,10 +26,8 @@ import (
 	"sigs.k8s.io/kwok/pkg/utils/slices"
 )
 
-var (
-	// ErrBrokenLinks is returned when there are broken links.
-	ErrBrokenLinks = fmt.Errorf("broken links dependency detected")
-)
+// ErrBrokenLinks is returned when there are broken links.
+var ErrBrokenLinks = fmt.Errorf("broken links dependency detected")
```
wzshiming, line 30: restore files that don't need changing.

hack/update-codegen.sh
```diff
@@ -49,6 +49,7 @@ function gen() {
     --input-dirs ./pkg/apis/v1alpha1/ \
     --input-dirs ./pkg/apis/config/v1alpha1/ \
     --input-dirs ./pkg/apis/action/v1alpha1/ \
+    --output-package sigs.k8s.io/kwok/pkg/apis \
```
wzshiming, line 52: Is your environment unable to execute this script?

wzshiming (commented):
Most LGTM and there are a few details that need tweaking.

pkg/kwokctl/runtime/kind/cluster.go
sivchari: filterDuplicatedExtraArgs overrides existing args using patch.ExtraArgs if args.Key is same.
So I look like that it's already designed. Am I misunderstanding ?

sivchari: What is indicated about expected behavior ?
Is it wrong that --extra-args patch the component if it's passed ? I think it's okay to change the expected behavior as a role of --extra-args. If am I misunderstanding, please let me know.

hack/update-codegen.sh
sivchari: Yes, I can't update, but this change might not be necessary because KwokctlConfiguration is not changed. So I restored.
Let's discuss this problem somewhere else.

pkg/kwokctl/runtime/binary/cluster.go
```diff
@@ -940,7 +940,7 @@ func (c *Cluster) Logs(ctx context.Context, name string, out io.Writer) error {
 		return nil
 	}
 
-	f, err := os.OpenFile(logs, os.O_RDONLY, 0640)
+	f, err := os.OpenFile(logs, os.O_RDONLY, 0o640)
```
wzshiming, line 943:
```suggestion
	f, err := os.OpenFile(logs, os.O_RDONLY, 0640)
```

I think it makes sense, but let's revert to it first, to keep the style consistent with the others.

We can change it all at once in the new PR with golangci to avoid these values being changed by different lint configurations.

pkg/kwokctl/cmd/create/cluster/cluster.go
```diff
@@ -145,6 +177,7 @@ func NewCommand(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringSliceVar(&flags.Options.EnableCRDs, "enable-crds", flags.Options.EnableCRDs, "List of CRDs to enable")
 	cmd.Flags().UintVar(&flags.Options.NodeLeaseDurationSeconds, "node-lease-duration-seconds", flags.Options.NodeLeaseDurationSeconds, "Duration of node lease in seconds")
 	cmd.Flags().Float64Var(&flags.Options.HeartbeatFactor, "heartbeat-factor", flags.Options.HeartbeatFactor, "Scale factor for all about heartbeat")
+	cmd.Flags().StringArrayVar(&flags.ExtraArgs, "extra-args", flags.ExtraArgs, extraArgsUsage)
```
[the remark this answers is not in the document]
sivchari: Is it enough to explain only format ?

pkg/kwokctl/runtime/binary/cluster.go
sivchari: Okay!

pkg/kwokctl/runtime/kind/cluster.go
sivchari: > that this thing is an array.

I may know what you mean. But if the same key is passed, both is written, isn't is ? For example

`name1=key1=value1` and `name1=key1=value2`

is equal to

```yaml
- name1:
   extraArgs:
     - key: key1
        value: value1
    - key: key1
       value: value2
```

right ?

If so, which value is used by kwok ? And I think this configure may confuse to developer that look it. What do you think ?

sivchari: How about that ?

wzshiming: I just noticed that Kind's configuration doesn't seem to allow keys to be duplicates? Maybe we should report the error and let the user decide?

wzshiming (commented):
Sorry for the late reply.

pkg/kwokctl/runtime/kind/cluster.go
sivchari: Sorry, I caught a cold, so late to respond.

I agree to file this problem to Kind community if a report is needed.
But I want to clarify why we allow keys to be duplicated.

If `name1=key1=value1` and `name1=key1=value2` are passed, what do you expect ?
And what do you suppose any usecase in kwok.

sivchari: I think it's different to pass duplicate elements to array and to process. So if we don't allow to pass same keys (as you referred to Kind), I think that we must make an error or filter same key.

sivchari: ping @wzshiming

wzshiming: ohh, I'm sorry if I wasn't clear and you misunderstood me, what I meant was to add a warning message to this PR while de-duplicating it to let the user know it's being ignored.

sivchari: Oh, I'm sorry that I misunderstood. I could got it.
I added log if the key is duplicated.

pkg/kwokctl/cmd/create/cluster/cluster.go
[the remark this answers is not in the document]
sivchari: I got it. Should I write a new article describing how to use `--extra-args` ?

pkg/kwokctl/runtime/kind/cluster.go
sivchari: I selected the warning to let users know about information. But, is it better to report as an error than warning ?
What do you think ?

sivchari:
@wzshiming 
Sorry for the late reply. I've done to revise some changes.

wzshiming (commented):
LGTM, and still need run `./hack/update-cmd-docs.sh` to update docs

sivchari:
@wzshiming 
I ran `hack/update-cmd-docs.sh`, thanks.

PR #1060, opened 2024-05-03
feature: added helm support

### LINKED ISSUE #1029: Support helm charts in kwok

It would be great if KWOK could support Helm charts. I had a brief discussion with @wzshiming on Slack, and here's a summary.

The main issue arises from the conflict with the go template's {{ and }}. Since Helm charts also use these as default values, KWOK would need to support modifying the default {{ and }}. This change is straightforward but requires modification at every occurrence.

To enable coexistence between the two, there are two options:

Integrate kustomize with Helm, as detailed here: https://helm.sh/docs/topics/advanced/#post-rendering
Integrate Helm with kustomize, as detailed here: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/helmcharts
Option 2 (integrating Helm into kustomize) appears to be the more feasible solution.

@dmitsh, what are your thoughts on this?

### Why is this needed?

Helm charts are widely used, making it easier to configure, create, and manage virtual clusters.

dmitsh:
#### What type of PR is this?
/kind feature

This PR adds support for deploying kwok controller via Helm chart.

Fixes #1029

NONE

```release-note
Support deployment of kwok controller and stages with helm charts
```

```docs

```


deployments/kwok/templates/stage-fast.yaml
wzshiming, line 1: I think there should be separate charts for deployment and stages,
the `deployments` should be renamed `charts` or `helm`.

WDYT?

wzshiming (commented):
It feels like there is a lot of duplication with kustomize.
Maybe we should get a way to convert the existing kustomize to a helm chart.


There is one more issue that needs to be discussed
Where is the release of the chart?
- [Chart Repository](https://helm.sh/docs/topics/chart_repository/)
  - GitHub Pages of this repo?
- [OCI registry](https://helm.sh/docs/topics/registries/)
  - registry.k8s.io/kwok/ (It's theoretically possible, but I haven't seen it done in other projects, and I don't know if it's supported by that CI.)?
  - docker.io ?

deployments/kwok/crds/crd.yaml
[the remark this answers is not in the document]
dmitsh: If the goal is to convert kustomize to helm, then the originals should be in the helm chart?

deployments/kwok/templates/stage-fast.yaml
dmitsh: Thanks! I have updated PR with these suggestions. PTAL.

yuanchen8911:
You should put the release note inside the `release-note `block in the PR description. Also, I'd suggest something like the following in the release note.

```release-note
Support helm chart
```

wzshiming:
I found a project that converts yaml to charts.
https://github.com/arttor/helmify

```yaml
IMAGE_PREFIX=registry.k8s.io/kwok GIT_TAG=v0.5.2 make manifests
cat artifacts/kwok.yaml | go run github.com/arttor/helmify/cmd/helmify@v0.4.12 -crd-dir ./charts/kwok
cat artifacts/stage-fast.yaml | sed -e 's#{{#{{`{{#g' -e 's#}}#}}`}}#g' | go run github.com/arttor/helmify/cmd/helmify@v0.4.12 -original-name ./charts/stage-fast
```

It feels good.

wzshiming:
Hi @dmitsh , do you have time to work on this PR? If not I'll continue this.

dmitsh:
> Hi @dmitsh , do you have time to work on this PR? If not I'll continue this.

Hi @wzshiming, I would be glad to continue working on this PR if we can agree on the specifics.
I believe creating a proper Helm chart is the best approach. Based on my experience, the "helmify" tool isn't always reliable and needs extra validation. 
If we maintain Helm chart, we will also remove the need for additional hack scripts and pipelines, and will enable better customization of deployment parameters.
I favor this approach. Please let me know if we have a consensus.

wzshiming:
Okay, I still want a script that verifies the consistency between the configurations maintained by Kustomize and the Helm chart. this includes elements such as CRD, RBAC, and the Stage.

dmitsh:
> Okay, I still want a script that verifies the consistency between the configurations maintained by Kustomize and the Helm chart. this includes elements such as CRD, RBAC, and the Stage.

It is possible to reuse CRD and RBAC, but Stage is problematic because of '{{

wzshiming:
https://github.com/wzshiming/kwok/commit/70d1e7f380a1a118b07a60ae95961f72e0607bdd#diff-b88e78af774470c6d5ae68b7ef27d74bc3d2354efca52fc02b37d37e51bb13c9

It is not that complicated, just a simple text replacement. Have a look to see if it is up to your expectations. 😄

wzshiming:
@dmitsh Hi
https://github.com/wzshiming/kwok/commit/2c212651bd4848fac6e4274b031a74f1baa18db9#diff-b88e78af774470c6d5ae68b7ef27d74bc3d2354efca52fc02b37d37e51bb13c9
Could you take a look at that, and leave your comments?

dmitsh:
Closing in favor of https://github.com/kubernetes-sigs/kwok/pull/1138

PR #1073, opened 2024-05-12
Fix replicas data type specified in kwokctl_scale.md

network-charles:
#### What type of PR is this?

/kind documentation

It corrects the data type specified for the `--replicas` flag. I changed it from `uint` to `int`.

```release-note
NONE
```


wzshiming:
This is generated by `./hack/update-cmd-docs.sh` with https://github.com/kubernetes-sigs/kwok/blob/f2ac0f927b1a6c41e7f77ce659b4f3833db6244d/pkg/kwokctl/cmd/scale/scale.go#L62

network-charles:
Thanks for clarifying, @wzshiming. I also looked up “uint” and I realized it's an unsigned integer that accepts non-negative values only.

PR #1104, opened 2024-05-22
adding tests for files

mohamedasifs123:
#### What type of PR is this?
kind feature

adding tests for files

Fixes #

```release-note
none
```

```docs
none
```

### DIFF
--- a/pkg/utils/file/compress_test.go
+++ b/pkg/utils/file/compress_test.go
@@ -1,9 +1,12 @@
 /*
 Copyright 2024 The Kubernetes Authors.
+
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
+
     http://www.apache.org/licenses/LICENSE-2.0
+
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
--- a/pkg/utils/file/compress_test.go
+++ b/pkg/utils/file/compress_test.go
@@ -54,14 +54,20 @@ func TestCompress(t *testing.T) {
 		if err != nil {
 			t.Fatalf("failed to write data: %v", err)
 		}
-		writer.Close()
+		if err := writer.Close(); err != nil {
+			t.Fatalf("failed to close: %v", err)
+		}
 
 		if test.expected {
 			gr, err := gzip.NewReader(&buf)
 			if err != nil {
 				t.Fatalf("expected gzip compression, got error: %v", err)
 			}
-			defer gr.Close()
+			defer func() {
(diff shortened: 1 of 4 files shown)


pkg/utils/file/compress_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"bytes"
+	"compress/gzip"
+	"io"
+	"testing"
+)
+
+func TestCompress(t *testing.T) {
+	tests := []struct {
+		name     string
+		ext      string
```
wzshiming, line 26:
```suggestion
		ext      string
		data     []byte
```

Don't hard-coded the data

pkg/utils/file/compress_test.go
```diff
@@ -0,0 +1,145 @@
...
+		} else if buf.String() != "hello world" {
+			t.Errorf("expected 'hello world', got %s", buf.String())
+		}
+	}
+}
+
+func TestDecompress(t *testing.T) {
+	tests := []struct {
+		name       string
+		compressed bool
```
wzshiming, line 87:
```suggestion
        data       []byte
		compressed bool
```

ditto

pkg/utils/file/compress_test.go
```diff
@@ -0,0 +1,145 @@
...
+		{
+
+			name:       "test.tgz",
+			compressed: true,
+		},
+		{
+
+			name:       "test.zip",
+			compressed: false,
+		},
```
wzshiming, line 107: add a case with gzip compressed data

pkg/utils/file/download_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"os"
+	"testing"
+
+	"sigs.k8s.io/kwok/pkg/utils/path"
+)
+
+func TestDownloadWithCacheAndExtract(t *testing.T) {
+	ctx := context.Background()
+	cacheDir := "/tmp/cache"
+	src := "http://example.com/file.tar.gz"
```
wzshiming, line 31: Don't depend on an external source, start a test server instead.

wzshiming (commented):
/ok-to-test

pkg/utils/file/download_test.go
mohamedasifs123: In test server where can I place the files?

pkg/utils/file/compress_test.go
mohamedasifs123: Shall I need to have separate data for all test cases?

pkg/utils/file/download_test.go
wzshiming: import "net/http/httptest" and use `httptest.NewServer()` to create a test server, and use `gzip` to compress the content as http response, and put the output in the tmp dir.

pkg/utils/file/compress_test.go
mohamedasifs123: I think The decompress func only support .gz iand .tgz .

PR #1018, opened 2024-04-03
Supports load config from kustomize

wzshiming:
#### What type of PR is this?

/kind feature

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
Supports load config from kustomize
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs
NONE
```

PR #1185, opened 2024-07-23
Fix a stage file path issue in demo

yuanchen8911:
#### What type of PR is this?

/kind bug

This PR fixes a non-existing stage file path in the file `stage-pod-general.demo`.

Fixes #

```release-note
Fix a broken stage file path in a demo.
```

```docs

```


demo/stages-pod-general.demo
```diff
@@ -1,5 +1,5 @@
 # Let's getting started with kwokctl!
-kwokctl -c ./stages/pod-general.yaml create cluster
+kwokctl -c ./kustomize/stage/pod/general/kustomization.yaml create cluster
```
wzshiming, line 2: Ah, now is unsupported use kustomization.yaml file as config, it will support in #1018

yuanchen8911: Any fix without the `kcustomization` support?

PR #1216, opened 2024-09-05
make GetUnusedPort concurrency safe

muff1nman:
#### What type of PR is this?
/kind bug

Kwok can be used nicely in go test with parallelism. However when running with -race, GetUnusedPort is flagged as a data race. Ancedotally, during our tests we have seen port contention and this might help resolve that.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/utils/net/unused_port.go
+++ b/pkg/utils/net/unused_port.go
@@ -20,17 +20,21 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"sync"
 
 	"sigs.k8s.io/kwok/pkg/utils/sets"
 )
 
 var (
-	errGetUnusedPort        = fmt.Errorf("unable to get an unused port")
+	errGetUnusedPort = fmt.Errorf("unable to get an unused port")
+	unusedPortSync   sync.Mutex
 	lastUsedPort     uint32 = 32767
 )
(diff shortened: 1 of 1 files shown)


wzshiming:
this function should never be used concurrently.

muff1nman:
@wzshiming are you saying that I shouldn't be using kwok with https://pkg.go.dev/testing#T.Parallel? I'd find that surprising considering how easy it is to use kwok in testing and with parallelism it makes the tests run very quickly.

muff1nman:
Actually I think there is more than just GetUnusedPorts that potentially is a problem.

wzshiming:
> Actually I think there is more than just GetUnusedPorts that potentially is a problem.

Yes, I know the kwokctl potentially has this problem, but there's no need for parallelism, right?

muff1nman:
The difference with t.Parallel is 3.5 minutes vs 23 minutes test runtime. So I'd say its very nice to have.

wzshiming:
Hello

I would like to verify what you said, and I have not found any way to do it in this codebase, how do you do it exactly, I would like to verify it.

muff1nman:
> I would like to verify what you said

Verify which part exactly?

PR #1260, opened 2024-11-02
Fix quantity scaling in cel calculations

### LINKED ISSUE #1259: Incorrect quantity calculations in CEL expressions
### How to use it?

- [X] kwok
- [ ] kwokctl --runtime=docker (default runtime)
- [ ] kwokctl --runtime=binary
- [ ] kwokctl --runtime=kind

### What happened?

I am currently simulating a large-scale cluster using Kwok to test Horizontal Pod Autoscaling (HPA). My setup involves simulating a lots of Pods with resource usage fluctuating randomly within a defined range.

While configuring the ClusterResourceUsage expression, I noticed an issue where performing calculations on Quantity leads to incorrect results.

To address this, I have submitted a simple PR to address this issue.

The result of Quantity in the CEL expression should be correct.

#### Normal Case

Directly specifying Quantity values in `ClusterResourceUsage` CRD for easy reproduction:

```yaml
kind: ClusterResourceUsage
metadata:
  name: usage-from-annotation
spec:
  usages:
  - usage:
      cpu:
        expression: |
          Quantity("1")
      memory:
        expression: |
          Quantity("2Gi")
```

Printing the resource usage of the Mock Pod:

```shell
$ kubectl top pod                   
NAME                        CPU(cores)   MEMORY(bytes)   
fake-pod-5c449fc8f7-2rt7z   1000m        2048Mi
```

When the expression in ClusterResourceUsage does not involve any operations on Quantity, the results are as expected.

#### Abnormal Case

Editing ClusterResourceUsage to use multiplication in the expressions, multiplying Quantity by 1. The expected result should remain consistent with the normal case.

```yaml
kind: ClusterResourceUsage
metadata:
  name: usage-from-annotation
spec:
  usages:
  - usage:
      cpu:
        expression: |
          Quantity("1") * 1.0
      memory:
        expression: |
          Quantity("2Gi") * 1.0
```

Printing the resource usage of the Mock Pod:

```shell
$ kubectl top pod
NAME                        CPU(cores)   MEMORY(bytes)   
fake-pod-5c449fc8f7-2rt7z   10001m       -8796Mi
```

Observed that the CPU usage is multiplied by a factor of ten, and the Memory value appears to have overflowed.

After consistently reproducing the problem, I read the Kwok source code and identified the issue in the `newQuantityFromFloat64` function of [`quantity.go`](https://github.com/kubernetes-sigs/kwok/blob/v0.6.1/pkg/utils/cel/quantity.go#L69):

```go
func newQuantityFromFloat64(v float64) Quantity {
    r := resource.NewScaledQuantity(int64(v * 10e9), resource.Nano)
    return NewQuantity(r)
}
```

The problem lies in the multiplication of the input `float64` value `v` by $10 \times 10^{9}$. Since `resource.Nano` represents $1 \times 10^{-9}$, this results in the function returning a value that is 10 times greater than expected.

Unfortunately, the unit tests did not catch this problem, as they also performed the incorrect calculation. In the `TestResourceEvaluation` function of [`evaluator_test.go`](https://github.com/kubernetes-sigs/kwok/blob/v0.6.1/pkg/kwok/metrics/evaluator_test.go#L68), the correct calculation should be:

```math
\begin{align*}
& ( \text{Quantity}("10m") + \text{Quantity}("90m") + \text{Quantity}("10m") + \text{Quantity}("10m") ) \times 1.5 \times 10 \\
& = \text{Quantity}("120m") \times 15 \\
& = \text{Quantity}("1800m") \\
& = 1.8
\end{align*}
```

However, the unit tests compared the CEL expression's result to 18.

<details>

```go
func TestResourceEvaluation(t *testing.T) {
	n := &corev1.Node{
		Status: corev1.NodeStatus{
			Allocatable: map[corev1.ResourceName]resource.Quantity{
				corev1.ResourceCPU: resource.MustParse("90m"),
			},
		},
	}
	p := &corev1.Pod{
		ObjectMeta: metav1.ObjectMeta{
			Annotations: map[string]string{
				"cpu_usage": "10m",
			},
		},
		Spec: corev1.PodSpec{
			Containers: []corev1.Container{
				{
					Resources: corev1.ResourceRequirements{
						Requests: map[corev1.ResourceName]resource.Quantity{
							corev1.ResourceCPU: resource.MustParse("10m"),
						},
					},
				},
			},
		},
	}

	exp := `( Quantity("10m") + Quantity(pod.metadata.annotations["cpu_usage"]) + node.status.allocatable["cpu"] + pod.spec.containers[0].resources.requests["cpu"] ) * 1.5 * 10`

	env, err := NewEnvironment(EnvironmentConfig{})
	if err != nil {
		t.Fatalf("failed to instantiate node Evaluator: %v", err)
	}

	eval, err := env.Compile(exp)
	if err != nil {
		t.Fatalf("failed to compile expression: %v", err)
	}

	actual, err := eval.EvaluateFloat64(context.Background(), Data{
		Node: n,
		Pod:  p,
	})
	if err != nil {
		t.Fatalf("evaluation failed: %v", err)
	}

	if actual != 18 {
		t.Errorf("expected %v, got %v", 18, actual)
	}
}
```

</details>

I have submitted a simple PR to address this issue.

### Kwok version

<details>

```console
$ kwok --version
0.6.1
```

</details>

### OS version

<details>

```console
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
$ uname -a
Linux act-1 6.8.0-47-generic #47~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Oct  2 16:16:55 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```
</details>

### LINKED ISSUE #1261: Overflow in quantity calculations for large values in CEL expressions
### How to use it?

- [X] kwok
- [ ] kwokctl --runtime=docker (default runtime)
- [ ] kwokctl --runtime=binary
- [ ] kwokctl --runtime=kind

### What happened?

This issue expands on a previously submitted issue #1259 regarding incorrect Quantity calculations in CEL expressions. 

While that issue was resolved, I encountered an additional limitation when handling large Quantity values, such as `20Gi`, which results in an overflow.

When carrying out operations on Quantity using CEL expressions, there should be no overflow in the result.

To reproduce the overflow issue, configure the `ClusterResourceUsage` as follows:

```yaml
kind: ClusterResourceUsage
metadata:
  name: large-quantity-test
spec:
  usages:
  - usage:
      memory:
        expression: |
          Quantity("20Gi") * 1.0
```

The expected result should handle 20Gi precisely. Nevertheless, during this procedure, the newQuantityFromFloat64 function in [quantity.go](https://github.com/kubernetes-sigs/kwok/blob/v0.6.1/pkg/utils/cel/quantity.go#L69) will carry out the following computation:

```go
func newQuantityFromFloat64(v float64) Quantity {
    r := resource.NewScaledQuantity(int64(v * 1e9), resource.Nano)
    return NewQuantity(r)
}
```

Here, `20Gi * 1e9` results in $21,474.836,480 * 1,000,000,000 = 21,474,836,480,000,000,000$, which exceeds the maximum value for int64 ($9,223,372,036,854,775,807$). This causes an overflow error, rendering the calculated quantity inaccurate.

```shell
$ kubectl top pod
NAME                        CPU(cores)   MEMORY(bytes)   
fake-pod-5c449fc8f7-2rt7z   1m       -8796Mi
```

Compared to issue #1259, this problem presents greater complexity. In the [`k8s.io/apimachinery` library](https://pkg.go.dev/k8s.io/apimachinery@v0.31.0/pkg/api/resource#Quantity.Mul), `Quantity` only supports multiplication by `int64` types:

```go
// Mul multiplies the provided y to the current value.
// It will return false if the result is inexact. Otherwise, it will return true.
func (q *Quantity) Mul(y int64) bool {
    q.s = ""
    if q.d.Dec == nil && q.i.Mul(y) {
        return true
    }
    return q.ToDec().d.Dec.Mul(q.d.Dec, inf.NewDec(y, inf.Scale(0))).UnscaledBig().IsInt64()
}
```

In the context of Kwok, users may want to dynamically calculate current resource usage based on a Pod's resource Requests or Limits. Examples include:

1. Periodically adjusting usage within a set range (e.g., from half the limits up to the full limits).
2. Performing random calculations based on the specified limits.

One workaround to avoid overflow in the `newQuantityFromFloat64` function is allowing users to directly specify the float64 Nano value in Annotations, thus bypassing `Quantity` altogether. However, this requires users to handle the calculations manually, adding to their mental load.

Another potential approach would be to trade some precision in cases of overflow, ensuring the result stays within the `int64` range.

After further discussion, I am very willing to participate in the implementation of the solution. Looking forward to collaborating on this improvement. 😊

### Kwok version

<details>

```console
$ kwok --version
0.6.1
```

</details>

### OS version

<details>

```console
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
$ uname -a
Linux act-1 6.8.0-47-generic #47~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Oct  2 16:16:55 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```
</details>

MondayCha:
#### What type of PR is this?

/kind bug

This Pull Request corrects inaccurate resource calculations in `ClusterResourceUsage` expressions by addressing a scaling issue in the `newQuantityFromFloat64` function, ensuring accurate resource usage results.

Fixes #1259
Fixes #1261

```release-note
Fix Quantity calculation
```

```docs
NONE
```

### DIFF
--- a/pkg/kwok/metrics/evaluator_test.go
+++ b/pkg/kwok/metrics/evaluator_test.go
@@ -18,6 +18,7 @@ package metrics
 
 import (
 	"context"
+	"math"
 	"testing"
 	"time"
 
@@ -112,7 +113,9 @@ func TestResourceEvaluation(t *testing.T) {
 		t.Fatalf("evaluation failed: %v", err)
 	}
 
-	if actual != 18 {
-		t.Errorf("expected %v, got %v", 18, actual)
+	const epsilon = 1e-9
+	expected := 1.8
+	if math.Abs(actual-expected) > epsilon {
+		t.Errorf("expected %v, got %v", expected, actual)
 	}
 }
--- a/pkg/utils/cel/funcs_test.go
+++ b/pkg/utils/cel/funcs_test.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cel
+
+import (
+	"testing"
+	"time"
+
+	"github.com/google/cel-go/common/types"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestTimeNow(t *testing.T) {
+	refVal := runAndCheckExpression(t, "Now()", types.TimestampType)
+
+	actual := refVal.Value().(time.Time)
+
+	if !expectedTime.Equal(actual) {
+		t.Errorf("expected %v, got %v", expectedTime, actual)
+	}
+}
+
+func TestMathRand(t *testing.T) {
+	refVal := runAndCheckExpression(t, "Rand()", types.DoubleType)
+	actual := refVal.Value().(float64)
+
+	if actual < 0 || actual > 1 {
+		t.Errorf("expected value between 0 and 1, got %v", refVal.Value())
+	}
+}
+
+func TestUnixSecond(t *testing.T) {
+	refVal := runAndCheckExpressionWithData(t, "UnixSecond(time)", map[string]any{
+		"time": expectedTime,
+	}, types.DoubleType)
+
+	actual := refVal.Value().(float64)
+	expected := float64(expectedTime.Unix())
+
+	compareFloat64(t, expected, actual)
+}
+
+func TestSinceSecond(t *testing.T) {
+	n := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			CreationTimestamp: metav1.Time{Time: expectedTime},
+		},
+	}
+	refVal := runAndCheckExpressionWithData(t, "SinceSecond(node)", map[string]any{
+		"node": n,
+	}, types.DoubleType)
+
+	actual := refVal.Value().(float64)
+	if actual < 0 {
+		t.Errorf("expected positive value, got %v", actual)
+	}
+}
--- a/pkg/utils/cel/quantity.go
+++ b/pkg/utils/cel/quantity.go
@@ -47,7 +47,7 @@ type Quantity struct {
 // NewQuantity creates a new Quantity
 func NewQuantity(q *resource.Quantity) Quantity {
 	if q == nil {
-		q = resource.NewScaledQuantity(0, resource.Nano)
+		q = resource.NewMilliQuantity(0, resource.DecimalSI)
 	}
 	return Quantity{Quantity: q}
 }
@@ -61,18 +61,18 @@ func NewQuantityFromString(s string) (Quantity, error) {
 	return NewQuantity(&r), nil
 }
 
-func newQuantityFromNanoInt64(v int64) Quantity {
-	r := resource.NewScaledQuantity(v, resource.Nano)
+func newQuantityFromMilliInt64(v int64) Quantity {
+	r := resource.NewMilliQuantity(v, resource.DecimalSI)
 	return NewQuantity(r)
 }
 
 func newQuantityFromFloat64(v float64) Quantity {
-	r := resource.NewScaledQuantity(int64(v*10e9), resource.Nano)
+	r := resource.NewMilliQuantity(int64(v*1e3), resource.DecimalSI)
 	return NewQuantity(r)
 }
 
-func (q Quantity) nano() int64 {
-	return q.Quantity.ScaledValue(resource.Nano)
+func (q Quantity) milli() int64 {
+	return q.Quantity.MilliValue()
 }
 
 func (q Quantity) float() float64 {
@@ -157,10 +157,10 @@ func (q Quantity) Divide(other ref.Val) ref.Val {
 	switch other.Type() {
 	case types.IntType:
 		otherInt := other.(types.Int)
-		return newQuantityFromNanoInt64(q.nano() / int64(otherInt))
+		return newQuantityFromMilliInt64(q.milli() / int64(otherInt))
 	case types.UintType:
 		otherUint := other.(types.Uint)
-		return newQuantityFromNanoInt64(q.nano() / int64(otherUint))
+		return newQuantityFromMilliInt64(q.milli() / int64(otherUint))
 	case types.DoubleType:
 		otherDouble := other.(types.Double)
 		return newQuantityFromFloat64(q.float() / float64(otherDouble))
@@ -174,10 +174,10 @@ func (q Quantity) Multiply(other ref.Val) ref.Val {
 	switch other.Type() {
 	case types.IntType:
 		otherInt := other.(types.Int)
-		return newQuantityFromNanoInt64(q.nano() * int64(otherInt))
+		return newQuantityFromMilliInt64(q.milli() * int64(otherInt))
 	case types.UintType:
 		otherUint := other.(types.Uint)
-		return newQuantityFromNanoInt64(q.nano() * int64(otherUint))
+		return newQuantityFromMilliInt64(q.milli() * int64(otherUint))
 	case types.DoubleType:
 		otherDouble := other.(types.Double)
 		return newQuantityFromFloat64(q.float() * float64(otherDouble))
--- a/pkg/utils/cel/quantity_test.go
+++ b/pkg/utils/cel/quantity_test.go
@@ -1,0 +1,297 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cel
+
+import (
+	"maps"
+	"math"
+	"testing"
+	"time"
+
+	"github.com/google/cel-go/common/types"
+	"github.com/google/cel-go/common/types/ref"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+var expectedTime = time.Date(2006, 1, 2, 15, 4, 5, 0, time.UTC)
+
+func TestQuantityCalculation(t *testing.T) {
+	tests := []struct {
+		name       string
+		expression string
+		expected   string
+	}{
+		{
+			name:       "parse tebibyte quantity",
+			expression: `Quantity("1Ti")`,
+			expected:   `1Ti`,
+		},
+		{
+			name:       "add tebibyte",
+			expression: `Quantity("1Ti") + Quantity("1Ti")`,
+			expected:   `2Ti`,
+		},
+		{
+			name:       "subtract gibibyte",
+			expression: `Quantity("1Ti") - Quantity("512Gi")`,
+			expected:   `512Gi`,
+		},
+		{
+			name:       "negate tebibyte",
+			expression: `-Quantity("1Ti")`,
+			expected:   `-1Ti`,
+		},
+		{
+			name:       "tebibyte multiply int",
+			expression: `Quantity("1Ti") * 2`,
+			expected:   `2Ti`,
+		},
+		{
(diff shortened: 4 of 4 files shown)


pkg/kwok/metrics/evaluator_test.go
```diff
@@ -66,53 +67,150 @@ func TestNodeEvaluation(t *testing.T) {
...
+			pod:        &corev1.Pod{},
+			expression: `Quantity("5m") / 2`,
+			expected:   0.002,
+		},
+		{
+			name:       "millicpu divide float",
+			node:       &corev1.Node{},
+			pod:        &corev1.Pod{},
+			expression: `Quantity("1m") / 0.1`,
+			expected:   0.01,
```
wzshiming, line 182: Move to ./pkg/utils/cel/quantity_test.go

pkg/utils/cel/quantity.go
```diff
@@ -47,7 +47,7 @@ type Quantity struct {
 // NewQuantity creates a new Quantity
 func NewQuantity(q *resource.Quantity) Quantity {
 	if q == nil {
-		q = resource.NewScaledQuantity(0, resource.Nano)
+		q = resource.NewMilliQuantity(0, resource.DecimalSI)
```
wzshiming, line 50: Use `NewQuantity` instead `NewMilliQuantity`, the Milli is still too small

pkg/utils/cel/quantity.go
```diff
@@ -61,18 +61,18 @@ func NewQuantityFromString(s string) (Quantity, error) {
...
 func newQuantityFromFloat64(v float64) Quantity {
-	r := resource.NewScaledQuantity(int64(v*10e9), resource.Nano)
+	r := resource.NewMilliQuantity(int64(v*1e3), resource.DecimalSI)
 	return NewQuantity(r)
 }
 
-func (q Quantity) nano() int64 {
-	return q.Quantity.ScaledValue(resource.Nano)
+func (q Quantity) milli() int64 {
+	return q.Quantity.MilliValue()
```
wzshiming, line 75: Remove this and use `.float()` instead

pkg/utils/cel/quantity.go
```diff
@@ -61,18 +61,18 @@ func NewQuantityFromString(s string) (Quantity, error) {
...
+func newQuantityFromMilliInt64(v int64) Quantity {
+	r := resource.NewMilliQuantity(v, resource.DecimalSI)
 	return NewQuantity(r)
 }
 
 func newQuantityFromFloat64(v float64) Quantity {
-	r := resource.NewScaledQuantity(int64(v*10e9), resource.Nano)
+	r := resource.NewMilliQuantity(int64(v*1e3), resource.DecimalSI)
 	return NewQuantity(r)
 }
```
wzshiming, line 72: Remove this

pkg/utils/cel/quantity.go
```diff
@@ -61,18 +61,18 @@ func NewQuantityFromString(s string) (Quantity, error) {
 	return NewQuantity(&r), nil
 }
 
-func newQuantityFromNanoInt64(v int64) Quantity {
-	r := resource.NewScaledQuantity(v, resource.Nano)
+func newQuantityFromMilliInt64(v int64) Quantity {
+	r := resource.NewMilliQuantity(v, resource.DecimalSI)
 	return NewQuantity(r)
 }
```
wzshiming, line 67: Use `NewQuantity` instead `NewMilliQuantity`, the Milli is still too small

pkg/utils/cel/quantity.go
MondayCha: > [!note]
> https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
> Kubernetes doesn't allow you to specify CPU resources with a precision finer than 1m or 0.001 CPU. To avoid accidentally using an invalid CPU quantity, it's useful to specify CPU units using the milliCPU form instead of the decimal form when using less than 1 CPU unit.

Milli is required because CPU resources are measured in millicores (m) as the basic unit.

pkg/kwok/metrics/evaluator_test.go
MondayCha: Ok, I will do it.

pkg/utils/cel/quantity_test.go
```diff
@@ -0,0 +1,284 @@
...
+	}{
+		{
+			name:       "parse tebibyte quantity",
+			expression: `Quantity("1Ti")`,
+			expected:   1099511627776,
+		},
+		{
+			name:       "add tebibyte",
+			expression: `Quantity("1Ti") + Quantity("1Ti")`,
+			expected:   2199023255552,
```
wzshiming, line 44: Use readable expected value
```suggestion
			expected:   `Quantity("2Ti")`,
```

pkg/utils/cel/quantity_test.go
```diff
@@ -0,0 +1,284 @@
...
+		Types:       types,
+		Conversions: conversions,
+		Methods:     methods,
+		Funcs:       funcs,
+		Vars: map[string]any{
+			"node":      corev1.Node{},
+			"pod":       corev1.Pod{},
+			"container": corev1.Container{},
+		},
+	})
```
wzshiming, line 272:
```suggestion
	funcs := maps.Clone(DefaultFuncs)
	funcs[nowName] = []any{func() time.Time {
		return time.Date(2006, 1, 2, 15, 4, 5, 0, time.UTC)
	}}
	env, err := NewEnvironment(EnvironmentConfig{
		Types:       DefaultTypes,
		Conversions: DefaultConversions,
		Methods:     FuncsToMethods(funcs),
		Funcs:       funcs,
		Vars:        data,
	})
```

Make Now return a fixed value, and explicitly compare its value in funcs_test.go.

pkg/utils/cel/funcs_test.go
```diff
@@ -0,0 +1,106 @@
...
+func TestUnixSecond(t *testing.T) {
+	n := &corev1.Node{
+		ObjectMeta: metav1.ObjectMeta{
+			// 2006-01-02 15:04:05
+			CreationTimestamp: metav1.Time{Time: time.Date(2006, 1, 2, 15, 4, 5, 0, time.UTC)},
+		},
+	}
+	refVal, err := runCELExpressionWithData(t, "UnixSecond(node.metadata.creationTimestamp)", map[string]any{
+		"node": n,
+	})
```
wzshiming, line 70:
```suggestion
	time := time.Date(2006, 1, 2, 15, 4, 5, 0, time.UTC)
	refVal, err := runCELExpressionWithData(t, "UnixSecond(time)", map[string]any{
		"time": time,
	})
```

pkg/utils/cel/quantity_test.go
MondayCha: Unfortunately, achieving this will likely require even more effort on my part. 

> [!note] 
> [Quantity](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/)
> When a Quantity is parsed from a string, it will remember the type of suffix it had, and will use the same type again when it is serialized.

In Kwok’s implementation, the Add and Subtract functions reuse the official library's methods, allowing them to retain the suffix upon deserialization. 

However, for multiplication and division, custom implementations are used, and this process loses the suffix information. As a result, when testing `Quantity("1Ti") * 2`, calling `String()` on the Quantity outputs `2199023255552` rather than `2Ti`. 

Do you have any suggestions on this situation?

wzshiming: Convert the expected value to milli, then compare it.

pkg/utils/cel/quantity_test.go
```diff
@@ -0,0 +1,297 @@
...
+		{
+			name:       "millicpu floor to zero",
+			expression: `Quantity("1m") * 0.5`,
+			expected:   `0`,
+		},
+		{
+			name:       "millicpu divide int",
+			expression: `Quantity("5m") / 2`,
+			expected:   `2m`,
+		},
```
wzshiming, line 112: At the moment the smallest CPU unit in the K8s is 1m, anything less than that is meaningless.

wzshiming (approved):
/approve
/lgtm

Thank you!

PR #1286, opened 2024-12-16
Fix daemonset deletion in general stage

MondayCha:
#### What type of PR is this?

/kind bug

Kwok provides a [WIP General Stage](https://github.com/kubernetes-sigs/kwok/tree/main/kustomize/stage/pod/general) configuration to simulate a more comprehensive Pod lifecycle. However, when DaemonSets with tolerations configured as shown below exist, the General Stage configuration prevents these DaemonSet Pods from being garbage collected.

```yaml
tolerations:
- operator: Exists
```

If users attempt to delete these Pods before deleting the node, the DaemonSet will automatically rebuild them. Therefore, for DaemonSets, users typically don't explicitly delete Pods but rather delete nodes, relying on [Kubernetes' GC mechanism](https://github.com/kubernetes/kubernetes/issues/111257#issuecomment-1192212834) to delete the Pods.

However, since the General Stage adds Finalizers to all Pods, it causes Kubernetes' native GC mechanism to fail, leaving Pods stuck in the Terminating state.

```yaml
finalizers:
  add:
  - value: 'kwok.x-k8s.io/fake'
```

To address this, this PR considers ignoring Finalizers for DaemonSet-type Pods. Additionally, it fixes the missing `startTime` field in pod-ready.yaml compared to the fast configuration.

NONE


wzshiming (commented):
I think we can remove all finalizers from the general stage

PR #1336, opened 2025-04-06
Support for configuring previous container logs

### LINKED ISSUE #1335: Support for previous container logs

Support simulation of logs from previously terminated containers. Currently, when the --previous flag is passed, the system returns logs from the current container instead of the previous one.

### Why is this needed?

This feature is important for accurately simulating real-world Kubernetes behavior, where the --previous flag is used to retrieve logs from a container’s prior instance (e.g., after a crash or restart).

MitchelHsu:
#### What type of PR is this?

/kind feature

- Supports configuration for terminated container logs of a pod, useful for simulating scenarios like CrashLoopBackOff.

Fixes #1335 

```release-note
Support for configuring previous container logs
```

```docs
NONE
```

### DIFF
--- a/pkg/kwok/server/debugging_logs.go
+++ b/pkg/kwok/server/debugging_logs.go
@@ -47,7 +47,11 @@ func (s *Server) GetContainerLogs(ctx context.Context, podName, podNamespace, co
 	}
 
 	opts := crilogs.NewLogOptions(logOptions, time.Now())
-	return readLogs(ctx, log.LogsFile, opts, stdout, stderr)
+	logsFile := log.LogsFile
+	if logOptions.Previous {
+		logsFile = log.PreviousLogsFile
+	}
+	return readLogs(ctx, logsFile, opts, stdout, stderr)
 }
 
 // getContainerLogs handles containerLogs request against the Kubelet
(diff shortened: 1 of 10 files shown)


pkg/kwok/server/debugging_logs.go
```diff
@@ -47,7 +47,11 @@ func (s *Server) GetContainerLogs(ctx context.Context, podName, podNamespace, co
 	}
 
 	opts := crilogs.NewLogOptions(logOptions, time.Now())
-	return readLogs(ctx, log.LogsFile, opts, stdout, stderr)
+	logsFile := log.LogsFile
+	if logOptions.Previous {
+		logsFile = log.PreviousLogsFile
+	}
+	return readLogs(ctx, logsFile, opts, stdout, stderr)
```
wzshiming, line 54: There may be a few more cases to consider
- The final logsFile is empty
- Both --follow and --previous are enabled at the same time

wzshiming (commented):
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kwok/1336/pull-kwok-verify-main/1909451340610801664

It needs to run `./hack/update-api-docs.sh` to update the generated docs.

PR #1367, opened 2025-05-02
[WIP] fix: patch node addresses with replace strategy

cartermckinnon:
#### What type of PR is this?

/kind bug

We need to replicate the logic here when patching node addresses, due to a mistake in the type annotations: https://github.com/kubernetes/kubernetes/blob/03a3c0c89161935bc2338f5754ebb1104f779af1/staging/src/k8s.io/component-helpers/node/util/status.go#L85-L134

Nodes with addresses will not be patched successfully in some configurations. When testing the AWS cloud-controller-manager (https://github.com/kubernetes/cloud-provider-aws/), the existing patch strategy seems to result in a duplicate entry in the addresses list, leading to a validation error on the server side:
```
ERROR failed to apply stage node=ip-192-168-12-227.us-west-2.compute.internal-clone-wdln-5109 stage=node-initialize err="failed to patch node ip-192-168-12-227.us-west-2.compute.internal-clone-wdln-5109: Node \"ip-192-168-12-227.us-west-2.compute.internal-clone-wdln-5109\" is invalid: status.addresses[14]: Duplicate value: core.NodeAddress{Type:\"InternalIP\", Address:\"192.168.15.183\"}"
```

This PR resolves the issue and the node is patched as expected. This patching logic is used by the cloud-provider node controller itself: https://github.com/kubernetes/cloud-provider/blob/b2843894dd41f0cc0c0606afcfa08ef2fec6b451/controllers/node/node_controller.go#L402

```release-note
Fixes an issue in patches computed for a Node's status.addresses.
```

```docs

```

### DIFF
--- a/pkg/utils/lifecycle/next.go
+++ b/pkg/utils/lifecycle/next.go
@@ -117,8 +117,20 @@ func computePatch(renderer gotpl.Renderer, resource any, patch internalversion.S
 		}
 		return patchData, types.MergePatchType, nil
 	case "":
-		switch resource.(type) {
-		case *corev1.Node, *corev1.Pod:
+		switch t := resource.(type) {
+		case *corev1.Node:
+			patchData, err := computeMergePatch(renderer, resource, patch.Root, patch.Template)
+			if err != nil {
+				return nil, "", err
+			}
+			if patch.Subresource == "status" {
+				patchData, err = fixupPatchForNodeStatusAddresses(patchData, t.Status.Addresses)
+				if err != nil {
+					return nil, "", err
+				}
+			}
+			return patchData, types.StrategicMergePatchType, nil
+		case *corev1.Pod:
 			patchData, err := computeMergePatch(renderer, resource, patch.Root, patch.Template)
 			if err != nil {
 				return nil, "", err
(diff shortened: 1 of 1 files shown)


cartermckinnon:
Well....maybe it's just me 😅

cartermckinnon:
Seems like the tests here are making a different assumption:
```
Error: fake-pod-f9895f994-2dc4v Pod: HostIP (172.18.0.2) is not equal to the IP on its Node ()
```

wzshiming:
It seems that this patch has changed the node addresses, resulting in the failure of features such as metric, exec, etc. provided by kwok for simulating Kubelet.

Could you please explain the intention of transplanting this patch?

I feel that it should be directly achievable by modifying the Stage.

https://github.com/kubernetes-sigs/kwok/blob/b8b888d874f625a19d99bbca743eb7c25fc4fa6f/kustomize/stage/node/heartbeat-with-lease/node-heartbeat-with-lease.yaml#L36-L48

wzshiming:
Can you have a try at squashing all the commits that make the commit history clear?

https://stackoverflow.com/questions/1657017/how-to-squash-all-git-commits-into-one

wzshiming:
Sorry for sending the wrong message.

Issue #1579, opened 2026-05-13
Support IPv6


wzshiming:
### What would you like to be added?

Add generator for ipv6, and update the documentation.

### Why is this needed?

Currently, kwok only supports ipv4 single stack, k8s supports ipv4 and ipv6 single or dual stack, we should follow suit.

PR #1657, opened 2026-06-10
fix: handle bracketed IPv6 hosts in kubeconfig output

immanuwell:
#### What type of PR is this?
/kind bug

Tiny fix for `kwokctl get kubeconfig --host` with bracketed IPv6 hosts.

Right now `--host '[::1]'` ends up as `https://[[::1]]:6443`, so the kubeconfig is broken. This trims the brackets before `net.JoinHostPort`, keeps the old host behavior, and adds a regression test. pretty small, but real.

Related to #1579

Repro:
1. `kwokctl create cluster`
2. `kwokctl get kubeconfig --host '[::1]'`
3. see `server: https://[[::1]]:6443`

After this change it stays `https://[::1]:6443`.

Checks:
- `go test ./pkg/kwokctl/cmd/get/kubeconfig -run TestModifyAddress -v`
- `go test ./...`

```release-note
Fix `kwokctl get kubeconfig --host` for bracketed IPv6 hosts without an explicit port.
```

```docs
NONE
```

### DIFF
--- a/pkg/kwokctl/cmd/get/kubeconfig/kubeconfig.go
+++ b/pkg/kwokctl/cmd/get/kubeconfig/kubeconfig.go
@@ -24,6 +24,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -207,7 +208,11 @@ func modifyAddress(origin string, address string) (string, error) {
 		if err != nil {
 			return "", err
 		}
-		u.Host = net.JoinHostPort(address, port)
+		u.Host = net.JoinHostPort(trimIPv6Brackets(address), port)
 	}
 	return u.String(), nil
+}
+
+func trimIPv6Brackets(host string) string {
+	return strings.TrimSuffix(strings.TrimPrefix(host, "["), "]")
 }
--- a/pkg/kwokctl/cmd/get/kubeconfig/kubeconfig_test.go
+++ b/pkg/kwokctl/cmd/get/kubeconfig/kubeconfig_test.go
@@ -109,3 +109,49 @@ func TestGetCurrentClusterAfterMinifyConfig(t *testing.T) {
 		t.Fatalf("getCurrentCluster() error = %v, wantErr %v", err, errCurrentClusterNotFound)
 	}
 }
+
+func TestModifyAddress(t *testing.T) {
+	tests := []struct {
+		name    string
+		origin  string
+		address string
+		want    string
+	}{
+		{
+			name:    "hostname without port keeps original port",
+			origin:  "https://127.0.0.1:6443",
+			address: "localhost",
+			want:    "https://localhost:6443",
+		},
+		{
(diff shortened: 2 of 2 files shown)


pkg/kwokctl/cmd/get/kubeconfig/kubeconfig.go
```diff
@@ -207,7 +208,11 @@ func modifyAddress(origin string, address string) (string, error) {
 		if err != nil {
 			return "", err
 		}
-		u.Host = net.JoinHostPort(address, port)
+		u.Host = net.JoinHostPort(trimIPv6Brackets(address), port)
```
wzshiming, line 211:
```golang
// JoinHostPort combines host and port into a network address of the
// form "host:port". If host contains a colon, as found in literal
// IPv6 addresses, then JoinHostPort returns "[host]:port".
//
// See func Dial for a description of the host and port parameters.
func JoinHostPort(host, port string) string {
	// We assume that host is a literal IPv6 address if host has
	// colons.
	if bytealg.IndexByteString(host, ':') >= 0 {
		return "[" + host + "]:" + port
	}
	return host + ":" + port
}
```

I think this should be contributed to the upstream Golang package

immanuwell: agreed, upstream would be better long-term
this PR is a small local workaround for a real broken case today: --host '[::1]' currently produces https://[[::1]]:6443
this keeps the kubeconfig valid and adds regression coverage

wzshiming: I feel very strange why the standard library doesn't support this, so I looked up the reason and found a related issue; one sentence from the maintainer convinced me.

> JoinHostPort joins a host and a port. [ipv6addr] is not a host.

https://github.com/golang/go/issues/26006#issuecomment-400077674

So I understand that [ipv6] is not a valid ipv6 or ipv6addr; it is only used to combine IPv6 and port together or to distinguish the IPv6 delimiter in a URL. Therefore, it should not have square brackets when it exists alone.


And ping is also not supported

```
> ping6 ::1
PING6(56=40+8+8 bytes) ::1 --> ::1
16 bytes from ::1, icmp_seq=0 hlim=64 time=0.141 ms
^C
--- ::1 ping6 statistics ---
1 packets transmitted, 1 packets received, 0.0% packet loss
round-trip min/avg/max/std-dev = 0.141/0.141/0.141/nan ms
> ping6 [::1]
zsh: no matches found: [::1]
```

immanuwell: makes sense, `[::1]` is url notation, not host, so upstream is the wrong place for this
the trim is just to handle what users naturally copy-paste from a url

immanuwell:
@wzshiming what is the current status of this PR?
are there any required actions from my side?

wzshiming:
I think this trim is not necessary, users can manually remove it.

immanuwell:
got it, so I am going to close this PR

PR #1699, opened 2026-06-30
fix: avoid shared temp file races in download cache

immanuwell:
#### What type of PR is this?

/kind bug
/kind failing-test

Fixes a race in the shared download cache.

On `main`, `go test ./...` can hit this when `test/e2e/kwokctl/benchmark` and `test/e2e/kwokctl/benchmark-hack` download the same etcd tarball at the same time. They both write to the same `.tmp` path, one wins the rename, the other blows up with `rename ... .tmp: no such file or directory`.

This patch gives each downloader its own temp file and returns the cached file if another worker already won the rename. Also adds a tight concurrent unit test, so this doesnt drift back later.

None

Repro on `main`:

1. Run `go test ./...`
2. Watch `test/e2e/kwokctl/benchmark` and `test/e2e/kwokctl/benchmark-hack` race on the same cache entry

Targeted checks I ran:

- `go test ./pkg/utils/file ./pkg/kwokctl/runtime ./pkg/kwokctl/components`
- `go test -race ./pkg/utils/file -run TestGetCacheOrDownloadConcurrent -count=50`

```release-note
NONE
```

```docs
```

### DIFF
--- a/pkg/utils/file/download.go
+++ b/pkg/utils/file/download.go
@@ -31,6 +31,7 @@ import (
 	"github.com/wzshiming/httpseek"
 
 	"sigs.k8s.io/kwok/pkg/log"
+	"sigs.k8s.io/kwok/pkg/utils/flock"
 	utilspath "sigs.k8s.io/kwok/pkg/utils/path"
 	"sigs.k8s.io/kwok/pkg/utils/progressbar"
 	"sigs.k8s.io/kwok/pkg/utils/version"
@@ -145,91 +146,172 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
 	}
 	switch u.Scheme {
 	case "http", "https":
-
-		logger := log.FromContext(ctx)
-		logger = logger.With(
-			"uri", src,
-		)
-		logger.Info("Download")
-
-		var transport = http.DefaultTransport
-		transport = httpseek.NewMustReaderTransport(transport, func(req *http.Request, retry int, err error) error {
-			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
-				return err
-			}
-			if retry > 10 {
-				return err
-			}
-			logger.Warn("Retry after 1s",
-				"err", err,
-				"retry", retry,
-			)
-			time.Sleep(time.Second)
-			return nil
-		})
-
-		if !quiet {
-			transport = progressbar.NewTransport(transport)
-		}
-
-		cli := &http.Client{
-			Transport: transport,
-		}
-		req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
-		if err != nil {
-			return "", err
-		}
-		req.Header.Set("User-Agent", version.DefaultUserAgent())
-		resp, err := cli.Do(req)
-		if err != nil {
-			return "", err
-		}
-
-		defer func() {
-			err = resp.Body.Close()
-			if err != nil {
-				logger.Error("Failed to close body of response",
-					"err", err,
-				)
-			}
-		}()
-
-		if resp.StatusCode != http.StatusOK {
-			return "", fmt.Errorf("%s: %s", u.String(), resp.Status)
-		}
-
-		err = os.MkdirAll(utilspath.Dir(cache), 0755)
-		if err != nil {
-			return "", err
-		}
-
-		d, err := os.OpenFile(cache+".tmp", os.O_CREATE|os.O_TRUNC|os.O_WRONLY, mode)
-		if err != nil {
-			return "", err
-		}
-
-		contentLength, err := io.Copy(d, resp.Body)
-		if err != nil {
-			_ = d.Close()
-			fmt.Println()
-			return "", err
-		}
-		err = d.Close()
-		if err != nil {
-			logger.Error("Failed to close file",
-				"err", err,
-			)
-		}
-		if resp.ContentLength != contentLength {
-			return "", fmt.Errorf("content length mismatch: %d != %d", resp.ContentLength, contentLength)
-		}
-
-		err = os.Rename(cache+".tmp", cache)
-		if err != nil {
-			return "", err
-		}
-		return cache, nil
+		return downloadHTTPWithCache(ctx, cache, src, u, mode, quiet)
 	default:
 		return src, nil
 	}
+}
+
+func lockIncompleteFile(ctx context.Context, path string) (*os.File, error) {
+	incompleteFile, err := flock.OpenFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	var lastUpdate = time.Now()
+	var offset int64
+	for {
+		err = flock.TryLock(incompleteFile)
+		if err == nil {
+			return incompleteFile, nil
+		}
+		if !errors.Is(err, flock.ErrLocked) {
+			_ = incompleteFile.Close()
+			return nil, fmt.Errorf("failed to lock download cache: %w", err)
+		}
+
+		off, err := incompleteFile.Seek(0, io.SeekEnd)
+		if err != nil {
+			_ = incompleteFile.Close()
+			return nil, err
+		}
+
+		if offset == off {
+			if time.Since(lastUpdate) > time.Minute {
+				_ = incompleteFile.Close()
+				return nil, fmt.Errorf("download is locked and not updated for more than 1 minute, please check the download process or remove the incomplete file: %s", path)
+			}
+		} else {
+			offset = off
+			lastUpdate = time.Now()
+			// TODO: Add progress bar to show the download progress of the other process.
+		}
+
+		select {
+		case <-ctx.Done():
+			_ = incompleteFile.Close()
+			return nil, ctx.Err()
+		case <-time.After(100 * time.Millisecond):
+		}
+	}
+}
+
+func downloadHTTPWithCache(ctx context.Context, cache, src string, u *url.URL, mode fs.FileMode, quiet bool) (string, error) {
+	err := os.MkdirAll(utilspath.Dir(cache), 0755)
+	if err != nil {
+		return "", err
+	}
+
+	tmp := cache + ".tmp"
+	incompleteFile, err := lockIncompleteFile(ctx, tmp)
+	if err != nil {
+		return "", err
+	}
+
+	logger := log.FromContext(ctx)
+	logger = logger.With("uri", src)
+	defer func() {
+		err := flock.Unlock(incompleteFile)
+		if err != nil {
+			logger.Error("Failed to unlock download cache", "path", tmp, "err", err)
+		}
+		err = incompleteFile.Close()
+		if err != nil {
+			logger.Error("Failed to close incomplete file", "path", tmp, "err", err)
+		}
+	}()
+
+	if _, err := os.Stat(cache); err == nil {
+		err = os.Remove(tmp)
+		if err != nil {
+			logger.Debug("Failed to remove incomplete file", "path", tmp, "err", err)
+		}
+		return cache, nil
+	}
+	logger.Info("Download")
+
+	var transport = http.DefaultTransport
+	transport = httpseek.NewMustReaderTransport(transport, func(req *http.Request, retry int, err error) error {
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			return err
+		}
+		if retry > 10 {
+			return err
+		}
+		logger.Warn("Retry after 1s",
+			"err", err,
+			"retry", retry,
+		)
+		time.Sleep(time.Second)
+		return nil
+	})
+
+	offset, err := incompleteFile.Seek(0, io.SeekEnd)
+	if err != nil {
+		return "", err
+	}
+	if !quiet {
+		transport = progressbar.NewTransportWithOffset(transport, uint64(offset))
+	}
+	cli := &http.Client{Transport: transport}
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+	if err != nil {
+		return "", err
+	}
+	req.Header.Set("User-Agent", version.DefaultUserAgent())
+	if offset > 0 {
+		req.Header.Set("Range", fmt.Sprintf("bytes=%d-", offset))
+	}
+	resp, err := cli.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		err := resp.Body.Close()
+		if err != nil {
+			logger.Error("Failed to close body of response", "err", err)
+		}
+	}()
+
+	if offset == 0 {
+		if resp.StatusCode != http.StatusOK {
+			return "", fmt.Errorf("GET %s: %s", u.String(), resp.Status)
+		}
+	} else {
+		switch resp.StatusCode {
+		case http.StatusPartialContent:
+			// OK
+		case http.StatusOK:
+			// Server ignored Range; restart from scratch.
+			if err := incompleteFile.Truncate(0); err != nil {
+				return "", err
+			}
+			if _, err := incompleteFile.Seek(0, io.SeekStart); err != nil {
+				return "", err
+			}
+		default:
+			return "", fmt.Errorf("GET Partial %s: %s", u.String(), resp.Status)
+		}
+	}
+
+	contentLength, err := io.Copy(incompleteFile, resp.Body)
+	if err != nil {
+		return "", err
+	}
+	if resp.ContentLength >= 0 && resp.ContentLength != contentLength {
+		return "", fmt.Errorf("content length mismatch: %d != %d", resp.ContentLength, contentLength)
+	}
+
+	err = os.Chmod(tmp, mode)
+	if err != nil {
+		return "", err
+	}
+	err = os.Rename(tmp, cache)
+	if err != nil {
+		if _, statErr := os.Stat(cache); statErr == nil {
+			return cache, nil
+		}
+		return "", err
+	}
+	return cache, nil
 }
--- a/pkg/utils/file/download_test.go
+++ b/pkg/utils/file/download_test.go
@@ -1,0 +1,147 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package file
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+func TestGetCacheOrDownloadConcurrent(t *testing.T) {
+	t.Parallel()
+
+	const body = "concurrent download cache content"
+	var requests atomic.Int32
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		requests.Add(1)
+		time.Sleep(50 * time.Millisecond)
+		_, _ = w.Write([]byte(body))
+	}))
+	defer srv.Close()
+
+	cacheDir := t.TempDir()
+
+	const workers = 8
+	var wg sync.WaitGroup
+	errCh := make(chan error, workers)
+	pathCh := make(chan string, workers)
+
+	wg.Add(workers)
+	for range workers {
+		go func() {
+			defer wg.Done()
+
+			path, err := getCacheOrDownload(context.Background(), cacheDir, srv.URL+"/archive.tgz", 0644, true)
+			if err != nil {
+				errCh <- err
+				return
+			}
+			pathCh <- path
+		}()
+	}
+
+	wg.Wait()
+	close(errCh)
+	close(pathCh)
+
+	for err := range errCh {
+		if err != nil {
+			t.Fatalf("getCacheOrDownload returned error: %v", err)
+		}
+	}
+
+	var cachePath string
+	for path := range pathCh {
+		if cachePath == "" {
+			cachePath = path
+			continue
+		}
+		if path != cachePath {
+			t.Fatalf("expected a shared cache path, got %q and %q", cachePath, path)
+		}
+	}
+
+	data, err := os.ReadFile(cachePath)
+	if err != nil {
+		t.Fatalf("failed to read cache file %q: %v", cachePath, err)
+	}
+	if string(data) != body {
+		t.Fatalf("unexpected cache contents: got %q, want %q", string(data), body)
+	}
+
+	if _, err := os.Stat(cachePath + ".tmp"); !errors.Is(err, os.ErrNotExist) {
+		t.Fatalf("expected incomplete download to be renamed, got err=%v", err)
+	}
+	if got := requests.Load(); got != 1 {
+		t.Fatalf("expected exactly one cold-start download, got %d", got)
+	}
+}
+
+func TestGetCacheOrDownloadResumesIncompleteFile(t *testing.T) {
+	t.Parallel()
+
+	const body = "partial download content"
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		http.ServeContent(w, r, "archive.tgz", time.Now(), strings.NewReader(body))
+	}))
+	defer srv.Close()
+
+	cacheDir := t.TempDir()
+	cachePath, err := getCachePath(cacheDir, srv.URL+"/archive.tgz")
+	if err != nil {
+		t.Fatalf("failed to build cache path: %v", err)
+	}
+	if err := os.MkdirAll(filepath.Dir(cachePath), 0755); err != nil {
+		t.Fatalf("failed to create cache directory: %v", err)
+	}
+	if err := os.WriteFile(cachePath+".incomplete", []byte(body[:8]), 0644); err != nil {
+		t.Fatalf("failed to create incomplete download: %v", err)
+	}
+
+	incomplete, err := os.ReadFile(cachePath + ".incomplete")
+	if err != nil {
+		t.Fatalf("failed to read incomplete download: %v", err)
+	}
+	if got, want := string(incomplete), body[:8]; got != want {
+		t.Fatalf("unexpected incomplete download: got %q, want %q", got, want)
+	}
+
+	path, err := getCacheOrDownload(context.Background(), cacheDir, srv.URL+"/archive.tgz", 0644, true)
+	if err != nil {
+		t.Fatalf("failed to resume download: %v", err)
+	}
+	data, err := os.ReadFile(path)
+	if err != nil {
+		t.Fatalf("failed to read resumed download: %v", err)
+	}
+	if got := string(data); got != body {
+		t.Fatalf("unexpected resumed download: got %q, want %q", got, body)
+	}
+	if _, err := os.Stat(cachePath + ".tmp"); !errors.Is(err, os.ErrNotExist) {
+		t.Fatalf("expected incomplete download to be renamed, got err=%v", err)
+	}
+}
--- a/pkg/utils/flock/flock.go
+++ b/pkg/utils/flock/flock.go
@@ -1,0 +1,45 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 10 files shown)


pkg/utils/file/download.go
```diff
@@ -203,10 +203,17 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
 		}
+		tmp := d.Name()
+		cleanup := true
+		defer func() {
+			if cleanup {
+				_ = os.Remove(tmp)
+			}
+		}()
 
 		contentLength, err := io.Copy(d, resp.Body)
```
wzshiming: If the file name is randomized, then the user interrupts the download, the temporary file will not be cleaned up.
I think to avoid this we should, for example, make the copy in a new goroutine, watch the ctx in the main process to see if it exits, and then clean up the temporary files when it does.

Actually, personally, I prefer to add file locks and support resume from breakpoints instead of random files

pkg/utils/file/download.go
```diff
@@ -224,10 +231,19 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
+		err = os.Chmod(tmp, mode)
+		if err != nil {
+			return "", err
+		}
+
+		err = os.Rename(tmp, cache)
 		if err != nil {
+			if _, statErr := os.Stat(cache); statErr == nil {
+				return cache, nil
+			}
```
wzshiming: The probability of this branch occurring is that two processes simultaneously download a file: one by the root user and another by a regular user. The root user would first complete the download, followed by the regular user. 

However, in reality, each user has a different default working directory, which makes this scenario unlikely to occur.

pkg/utils/file/download.go
immanuwell: got it, the updated code still uses per-download temp files to avoid the shared `.tmp` race, but it now also watches `ctx.Done()` and closes the response/file handles on cancellation so the deferred cleanup removes the temp file promptly

pkg/utils/file/download.go
immanuwell: agreed, the root-vs-non-root example is not the main case

I kept this as a defensive guard for concurrent publishers of the same cache entry. 
The race is on the explicit cache path, not the process working directory, and on non-Unix platforms `os.Rename` is less predictable here. 
If `cache` already exists at that point, returning it is still the right result

immanuwell:
@wzshiming addressed your suggestions, PTAL

pkg/utils/file/download.go
wzshiming: When multiple processes are started cold, the same file is downloaded twice. I think we should fix this problem completely.

We can introduce [file lock](https://github.com/kubernetes-sigs/etcd-manager/blob/84b6f946e53a0a6ab3530c96db5a9154913d4d4c/pkg/locking/flock.go).
- I'm not sure if the lock will be automatically released when program ends.

The first for downloading with ".incomplete" ext, while the incoming ones wait for first to be completion.


For files that were interrupted during download, We can [set download offset for resume progress](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Range), if the server does not support it, revret to full download. and also need to set the starting of the progress bar.

immanuwell: duplicate-download problem fixed: the latest push adds a per-cache-key file lock, writes through `<cache>.incomplete`, and makes concurrent
  cold-start callers wait for the first downloader and then reuse the finished cache file

interrupted/canceled downloads clean up the incomplete file

immanuwell:
@wzshiming fixed duplicate-download problem, WDYT?

pkg/utils/file/download_lock_unix.go
```diff
@@ -0,0 +1,70 @@
...
+	"time"
+
+	"golang.org/x/sys/unix"
+)
+
+const cacheLockRetryInterval = 100 * time.Millisecond
+
+type cacheLockHandle struct {
+	file *os.File
+}
```
wzshiming: This part of the two implementations is the same and is extracted and placed in a separate lock file.

pkg/utils/file/download_lock_unix.go
```diff
@@ -0,0 +1,70 @@
...
+
+package file
+
+import (
+	"context"
+	"errors"
+	"os"
+	"time"
+
+	"golang.org/x/sys/unix"
```
wzshiming: Switch to using syscall.Flock

pkg/utils/file/download.go
```diff
@@ -198,15 +220,22 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
 			return "", fmt.Errorf("%s: %s", u.String(), resp.Status)
 		}
 
-		err = os.MkdirAll(utilspath.Dir(cache), 0755)
-		if err != nil {
+		tmp := cache + ".incomplete"
+		err = os.Remove(tmp)
```
wzshiming, line 225: If the last downloaded file exists we should resume the download progress instead of discarding it.

pkg/utils/flock/flock_unix.go
wzshiming, line 1: Create a new pkg/utils/flock for future reuse

pkg/utils/file/download_lock_unix.go
immanuwell: done

pkg/utils/file/download_lock_unix.go
immanuwell: got it, fixed

pkg/utils/file/download.go
immanuwell: done

pkg/utils/flock/flock_unix.go
immanuwell: done

immanuwell:
@wzshiming done, addressed all 4 latest suggestions

pkg/utils/flock/flock_unix.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"context"
+	"errors"
+	"os"
+	"syscall"
+	"time"
+)
+
+const retryInterval = 100 * time.Millisecond
```
wzshiming: Move to flock.go

pkg/utils/flock/flock_unix.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Lock acquires an exclusive advisory lock on path.
+func Lock(ctx context.Context, path string) (*LockHandle, error) {
```
wzshiming: Add a timeout for when a .lock file is left behind, so that the alarm quits instead of staying stuck.

pkg/utils/flock/flock.go
wzshiming, line 1: We should add unit for flock to make sure it works

pkg/utils/progressbar/progressbar.go
```diff
@@ -55,14 +55,16 @@ func (p *transport) RoundTrip(req *http.Request) (*http.Response, error) {
 		return resp, nil
 	}
```
wzshiming, line 64: Only needed when the Status Code is http.StatusPartialContent (206)

pkg/utils/file/download.go
wzshiming: Move after getting the offset, passing the offset in as the start of the progress bar.

pkg/utils/file/download.go
```diff
@@ -175,57 +198,80 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
+		var offset int64
+		if info, statErr := os.Stat(tmp); statErr == nil {
+			offset = info.Size()
+		} else if !errors.Is(statErr, os.ErrNotExist) {
+			return "", statErr
 		}
 
-		defer func() {
-			err = resp.Body.Close()
+		for {
```
wzshiming: This part of the loop attempt is unnecessary, and is already done in `transport = httpseek.NewMustReaderTransport(transport, ...`

pkg/utils/file/download.go
```diff
@@ -145,6 +146,28 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
+				log.FromContext(ctx).Error("Failed to unlock download cache",
+					"cache", cache,
+					"err", unlockErr,
+				)
+			}
+		}()
+
+		if _, err := os.Stat(cache); err == nil {
+			return cache, nil
+		}
```
wzshiming: Move to before getting the flock, when the file already exists, that don't need lock.

pkg/utils/progressbar/reader.go
wzshiming, line 1: We should add NewReaderWithOffset after NewReader

wzshiming (changes requested):

pkg/utils/flock/flock_unix.go
immanuwell: done

pkg/utils/flock/flock_unix.go
immanuwell: done

pkg/utils/flock/flock.go
immanuwell: got it, done

pkg/utils/progressbar/progressbar.go
immanuwell: done, offset is applied only for 206 partial content

pkg/utils/file/download.go
immanuwell: moved incomplete-file offset discovery before lock/transport setup

pkg/utils/file/download.go
immanuwell: removed the redundant request retry loop

pkg/utils/file/download.go
immanuwell: done

pkg/utils/progressbar/reader.go
immanuwell: added it

immanuwell:
@wzshiming addressed all your suggestions, PTAL

pkg/utils/file/download.go
```diff
@@ -180,54 +210,58 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
-			if err != nil {
-				logger.Error("Failed to close body of response",
-					"err", err,
-				)
-			}
-		}()
-
-		if resp.StatusCode != http.StatusOK {
+		if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusPartialContent {
+			_ = resp.Body.Close()
```
wzshiming: Why remove this defer closing body, I think the original logic was good

pkg/utils/file/download.go
wzshiming: Move to before `log.FromContext(ctx).Error("Failed to unlock download cache",`, make it use this.

pkg/utils/file/download_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	const body = "partial download content"
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if got, want := r.Header.Get("Range"), "bytes=8-"; got != want {
+			t.Errorf("unexpected range header: got %q, want %q", got, want)
+		}
+		w.Header().Set("Content-Length", "16")
+		w.Header().Set("Content-Range", "bytes 8-23/24")
+		w.WriteHeader(http.StatusPartialContent)
+		_, _ = w.Write([]byte(body[8:]))
```
wzshiming:
```suggestion
		http.ServeContent(w, r, "archive.tgz", time.Now(), strings.NewReader(body))
```

pkg/utils/file/download.go
```diff
@@ -145,6 +146,35 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
+		}
+		tmp := cache + ".incomplete"
+		var offset int64
+		if info, statErr := os.Stat(tmp); statErr == nil {
+			offset = info.Size()
+		} else if !errors.Is(statErr, os.ErrNotExist) {
+			return "", statErr
+		}
+
+		lock, err := flock.Lock(ctx, cache+".lock")
```
wzshiming:
```suggestion
		lock, err := flock.Lock(ctx, tmp)
```

We can use incomplete file as a lock file as well.

pkg/utils/flock/flock.go
```diff
@@ -0,0 +1,50 @@
...
+type LockHandle struct {
+	file *os.File
+}
+
+// Lock acquires an exclusive advisory lock on path.
+func Lock(ctx context.Context, path string) (*LockHandle, error) {
+	ctx, cancel := context.WithTimeout(ctx, lockTimeout)
+	defer cancel()
+	return lock(ctx, path)
+}
```
wzshiming:
```suggestion
func Lock(ctx context.Context, path string) (*LockHandle, error) {
	ctx, cancel := context.WithTimeout(ctx, lockTimeout)
	defer cancel()
	handle, err := lock(ctx, path)
	if err != nil {
		return nil, fmt.Errorf("failed to acquire lock on %s: %w", path, err)
	}
	return handle, nil
}
```

pkg/utils/file/download.go
```diff
@@ -180,54 +210,58 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
-			logger.Error("Failed to close file",
-				"err", err,
-			)
+		if closeErr != nil {
+			return "", closeErr
 		}
-		if resp.ContentLength != contentLength {
+		if bodyCloseErr != nil {
+			logger.Error("Failed to close body of response", "err", bodyCloseErr)
+		}
```
wzshiming:
```suggestion

		tmp := cache + ".incomplete"

		incompleteFile, err := os.OpenFile(tmp, os.O_CREATE|os.O_RDWR, 0644)
		if err != nil {
			return "", err
		}

		logger := log.FromContext(ctx)
		logger = logger.With(
			"uri", src,
		)
		logger.Info("Download")

		defer func() {
			err := incompleteFile.Close()
			if err != nil {
				logger.Error("Failed to close incomplete file",
					"path", tmp,
					"err", err,
				)
			}
		}()

		var lastUpdate = time.Now()
		var offset int64

		for {
			err = flock.Lock(incompleteFile)
			if err == nil {
				break
			}

			if !errors.Is(err, flock.ErrLocked) {
				return "", fmt.Errorf("failed to lock download cache: %w", err)
			}

			off, err := incompleteFile.Seek(0, io.SeekEnd)
			if err != nil {
				return "", err
			}

			if offset == off && time.Since(lastUpdate) > time.Minute {
				return "", fmt.Errorf("download is locked and not updated for more than 1 minute, please check the download process or remove the incomplete file: %s", tmp)
			}

			// TODO: Show the download progress

			offset = off
			lastUpdate = time.Now()

			select {
			case <-ctx.Done():
				return "", ctx.Err()
			case <-time.After(100 * time.Millisecond):
			}
		}

		if _, err := os.Stat(cache); err == nil {
			err = flock.Unlock(incompleteFile)
			if err != nil {
				logger.Error("Failed to unlock download cache",
					"path", tmp,
					"err", err,
				)
			}

			err = os.Remove(tmp)
			if err != nil {
				logger.Debug("Failed to remove incomplete file",
					"path", tmp,
					"err", err,
				)
			}

			return cache, nil
		}

		defer func() {
			err := flock.Unlock(incompleteFile)
			if err != nil {
				logger.Error("Failed to unlock download cache",
					"path", cache,
					"err", err,
				)
			}
		}()

		var transport = http.DefaultTransport
		transport = httpseek.NewMustReaderTransport(transport, func(req *http.Request, retry int, err error) error {
			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
				return err
			}
			if retry > 10 {
				return err
			}
			logger.Warn("Retry after 1s",
				"err", err,
				"retry", retry,
			)
			time.Sleep(time.Second)
			return nil
		})

		offset, err = incompleteFile.Seek(0, io.SeekEnd)
		if err != nil {
			return "", err
		}

		if !quiet {
			transport = progressbar.NewTransportWithOffset(transport, uint64(offset))
		}

		cli := &http.Client{
			Transport: transport,
		}
		req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
		if err != nil {
			return "", err
		}
		req.Header.Set("User-Agent", version.DefaultUserAgent())
		if offset > 0 {
			req.Header.Set("Range", fmt.Sprintf("bytes=%d-", offset))
		}
		resp, err := cli.Do(req)
		if err != nil {
			return "", err
		}
		defer func() {
			err = resp.Body.Close()
			if err != nil {
				logger.Error("Failed to close body of response",
					"err", err,
				)
			}
		}()

		if offset == 0 {
			if resp.StatusCode != http.StatusOK {
				return "", fmt.Errorf("GET %s: %s", u.String(), resp.Status)
			}
		} else {
			if resp.StatusCode != http.StatusPartialContent {
				return "", fmt.Errorf("GET Partial %s: %s", u.String(), resp.Status)
			}
		}

		contentLength, copyErr := io.Copy(incompleteFile, resp.Body)
		if copyErr != nil {
			return "", copyErr
		}
```

pkg/utils/flock/flock.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Unlock releases the advisory lock.
+func (l *LockHandle) Unlock() error {
+	err := unlock(l.file)
+	closeErr := l.file.Close()
+	if err != nil {
+		return err
+	}
+	return closeErr
+}
```
wzshiming:
```suggestion
var (
	ErrLocked = fmt.Errorf("file is already locked")
)

// Lock acquires an advisory lock on the file.
func Lock(f *os.File) error {
	return lock(f)
}

// Unlock releases the advisory lock on the file.
func Unlock(f *os.File) error {
	return unlock(f)
}
```

pkg/utils/flock/flock_unix.go
```diff
@@ -0,0 +1,56 @@
...
+		timer := time.NewTimer(retryInterval)
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			_ = f.Close()
+			return nil, ctx.Err()
+		case <-timer.C:
+		}
+	}
+}
```
wzshiming:
```suggestion
func lock(f *os.File) error {
	err := syscall.Flock(int(f.Fd()), syscall.LOCK_EX|syscall.LOCK_NB)
	if err == nil {
		return nil
	}
	if !errors.Is(err, syscall.EWOULDBLOCK) && !errors.Is(err, syscall.EAGAIN) {
		return err
	}
	return ErrLocked
}
```

pkg/utils/flock/flock_windows.go
```diff
@@ -0,0 +1,59 @@
...
+		timer := time.NewTimer(retryInterval)
+		select {
+		case <-ctx.Done():
+			timer.Stop()
+			_ = f.Close()
+			return nil, ctx.Err()
+		case <-timer.C:
+		}
+	}
+}
```
wzshiming:
```suggestion
func lock(f *os.File) error {
	var overlapped windows.Overlapped
	err := windows.LockFileEx(windows.Handle(f.Fd()), windows.LOCKFILE_EXCLUSIVE_LOCK|windows.LOCKFILE_FAIL_IMMEDIATELY, 0, 1, 0, &overlapped)
	if err == nil {
		return nil
	}
	if !errors.Is(err, windows.ERROR_LOCK_VIOLATION) {
		return err
	}
	return ErrLocked
}
```

wzshiming (commented):
I've reorganized the code. What do you think?

wzshiming:
When two processes download a tar file concurrently, an error is thrown here.
https://github.com/kubernetes-sigs/kwok/blob/a89db6c2d4db3d01327b3a0f7731db6a9716cfa1/pkg/utils/file/download.go#L65-L68

Ignoring this error is a good way to get around it.
In the future, tar files should be handled more gracefully, it can be left alone in this PR for now.

wzshiming:
Started both clusters concurrently, also got an error in the kubeconfig modification

```
ERROR Failed to add context to kubeconfig err="failed to modify kubeconfig file: open /Users/zsm/.kube/config.lock: file exists"
```

it can be left alone in this PR for now.

pkg/utils/file/download.go
immanuwell: got it, fixed in the latest push here

pkg/utils/file/download.go
immanuwell: done

pkg/utils/file/download_test.go
immanuwell: applied this in the latest commit

pkg/utils/file/download.go
immanuwell: applied this in the latest commit

pkg/utils/flock/flock.go
immanuwell: applied this in the latest commit

pkg/utils/file/download.go
immanuwell: applied this in the latest commit

pkg/utils/flock/flock_windows.go
immanuwell: applied this in the latest commit

pkg/utils/flock/flock_unix.go
immanuwell: applied this in the latest commit

pkg/utils/flock/flock.go
immanuwell: applied this in the latest commit

immanuwell:
@wzshiming applied your suggested changes (and thank you for taking the time to reorganize this, it would have taken way longer otherwise))

pkg/utils/file/download.go
```diff
@@ -145,91 +146,186 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
+	}
+	if resp.ContentLength >= 0 && resp.ContentLength != contentLength {
+		return "", fmt.Errorf("content length mismatch: %d != %d", resp.ContentLength, contentLength)
+	}
+
+	err = os.Chmod(tmp, mode)
+	if err != nil {
+		return "", err
+	}
+	err = flock.Unlock(incompleteFile)
```
wzshiming: This shouldn't be unlock before moving the file, which will cause other processes that are waiting to get it and try to resume downloading the file. and, since the file has been downloaded in its entirety, the download offset will exceed the file size. Different servers may implement it differently, possibly responding with an error message

pkg/utils/file/download.go
```diff
@@ -145,91 +146,186 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
+	}
+	defer func() {
+		if !locked {
+			return
 		}
-		err = d.Close()
+		err := flock.Unlock(incompleteFile)
 		if err != nil {
-			logger.Error("Failed to close file",
+			log.FromContext(ctx).Error("Failed to unlock download cache",
```
wzshiming:
```suggestion
			logger.Error("Failed to unlock download cache",
```

pkg/utils/flock/flock_windows.go
```diff
@@ -14,25 +16,28 @@ See the License for the specific language governing permissions and
...
-		consts.ComponentKubeScheduler,
-		consts.ComponentKwokController,
-		consts.ComponentPrometheus,
-		consts.ComponentMetricsServer,
-		consts.ComponentSchedulerPlugins,
-	).Feature()
-	testEnv.Test(t, f0)
+func lock(f *os.File) error {
+	var overlapped windows.Overlapped
+	err := windows.LockFileEx(windows.Handle(f.Fd()), windows.LOCKFILE_EXCLUSIVE_LOCK|windows.LOCKFILE_FAIL_IMMEDIATELY, 0, 1, 0, &overlapped)
```
wzshiming, line 62:
```suggestion
	err := windows.LockFileEx(windows.Handle(f.Fd()), windows.LOCKFILE_FAIL_IMMEDIATELY, 0, 1, 0, &overlapped)
```

I was concerned about the inconsistent behavior between Linux and Windows. I looked up the documentation and found that if the Windows flock is set to `windows.LOCKFILE_EXCLUSIVE_LOCK`, other processes will not be able to read the contents of the file while it is locked.

https://learn.microsoft.com/en-us/previous-versions/windows/embedded/ms891385(v=msdn.10)

pkg/utils/flock/flock.go
```diff
@@ -14,24 +14,24 @@ See the License for the specific language governing permissions and
...
-		consts.ComponentKubeScheduler,
-		consts.ComponentKwokController,
-		consts.ComponentPrometheus,
-		consts.ComponentMetricsServer,
-	).Feature()
-	testEnv.Test(t, f0)
+// Lock acquires an advisory lock on the file.
+func Lock(f *os.File) error {
+	return lock(f)
+}
```
wzshiming:
```suggestion
// TryLock attempts to acquire an exclusive file lock on the open file f
// without blocking. Returns ErrLocked if the lock is held by another
// process.
func TryLock(f *os.File) error {
	return tryLock(f)
}
```

I found that flock actually has an option to wait for unlock before returning, our implementation is more TryLock than Lock, rename it to make it clearer.

pkg/utils/file/download.go
immanuwell: done

pkg/utils/file/download.go
immanuwell: done

pkg/utils/flock/flock_windows.go
immanuwell: done

pkg/utils/flock/flock.go
immanuwell: done

immanuwell:
@wzshiming addressed 3 latest suggestions, PTAL

pkg/utils/file/download.go
```diff
@@ -145,91 +146,158 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
+}
+
+func downloadHTTPWithCache(ctx context.Context, cache, src string, u *url.URL, mode fs.FileMode, quiet bool) (string, error) {
+	err := os.MkdirAll(utilspath.Dir(cache), 0755)
+	if err != nil {
+		return "", err
+	}
+	if _, err := os.Stat(cache); err == nil {
+		return cache, nil
+	}
```
wzshiming:
```suggestion
```

This check is duplicated in getCacheOrDownload.

pkg/utils/file/download.go
```diff
@@ -145,91 +146,158 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
-
-		d, err := os.OpenFile(cache+".tmp", os.O_CREATE|os.O_TRUNC|os.O_WRONLY, mode)
+	}()
+	defer func() {
+		err := flock.Unlock(incompleteFile)
 		if err != nil {
-			return "", err
+			logger.Error("Failed to unlock download cache", "path", tmp, "err", err)
 		}
+	}()
```
wzshiming:
```suggestion
	defer func() {
		err := flock.Unlock(incompleteFile)
if err != nil {
			logger.Error("Failed to unlock download cache", "path", tmp, "err", err)
}
		err = incompleteFile.Close()
if err != nil {
			logger.Error("Failed to close incomplete file", "path", tmp, "err", err)
}
	}()
```

wzshiming (commented):
> {"function":"main.main","file":"D:/a/kwok/kwok/cmd/kwokctl/main.go","line":56},"msg":"Execute exit","err":"write D:/a/kwok/kwok/workdir/cache/https/github.com/etcd-io/etcd/releases/download/v3.6.10/etcd-v3.6.10-windows-amd64.zip.incomplete: The process cannot access the file because another process has locked a portion of the file."}

https://github.com/kubernetes-sigs/kwok/actions/runs/30885171847/job/91956724975?pr=1699

In Windows, the behavior of flock seems to be a bit different.

pkg/utils/file/download.go
immanuwell: done

immanuwell:
@wzshiming applied your suggestions

pkg/utils/file/download.go
```diff
@@ -145,91 +146,155 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
+			return nil, err
 		}
-		req.Header.Set("User-Agent", version.DefaultUserAgent())
-		resp, err := cli.Do(req)
-		if err != nil {
-			return "", err
+		if offset == off && time.Since(lastUpdate) > time.Minute {
+			_ = incompleteFile.Close()
+			return nil, fmt.Errorf("download is locked and not updated for more than 1 minute, please check the download process or remove the incomplete file: %s", path)
 		}
```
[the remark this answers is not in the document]
wzshiming: The lastUpdate should only be updated when the file size changes.

pkg/utils/file/download.go
```diff
@@ -145,91 +146,155 @@ func getCacheOrDownload(ctx context.Context, cacheDir, src string, mode fs.FileM
...
-		return src, nil
+	}()
+
+	if offset == 0 {
+		if resp.StatusCode != http.StatusOK {
+			return "", fmt.Errorf("GET %s: %s", u.String(), resp.Status)
+		}
+	} else if resp.StatusCode != http.StatusPartialContent {
+		return "", fmt.Errorf("GET Partial %s: %s", u.String(), resp.Status)
+	}
```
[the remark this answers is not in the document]
wzshiming: It makes sense.

pkg/utils/file/download.go
wzshiming: it has not been solved. Merge Unlock and Close into a defer function

PR #1722, opened 2026-07-21
support bare boolean flags in componentsPatches extraArgs for kwokctl

Priyankasaggu11929:
#### What type of PR is this?

/kind feature

Currently, [applyComponentPatch](https://github.com/kubernetes-sigs/kwok/blob/a89db6c2d4db3d01327b3a0f7731db6a9716cfa1/pkg/kwokctl/runtime/util.go#L113-L127) function formats components arguments as `--key=value`.

Which means if i have to pass a bare boolean flag to components like Prometheus which accept boolean flags (as `--web.enable-lifecycle` and not `--web.enable-lifecycle=true`), it is not possible today.

I have tried with passing an empty value, but that gets rendered as `--web.enable-lifecycle=` (see the trailing `=`.

PR is proposing the approach followed by [prometheus-operator](https://github.com/prometheus-operator/prometheus-operator/blob/33334e1ec1756f526aae85dc4690f7a603883a8f/pkg/operator/argument.go#L70-L76) to handle this scenario.

---

**Note:**  we are in process of implementing scale tests in the [k-sigs/node-readiness-controller](https://github.com/kubernetes-sigs/node-readiness-controller/) project using kwok.

As part of our setup, we have a requirement to configure the embedded prometheus (in kwok) to also scrape NRC metrics, which requires reloading Prometheus using the [reload API endpoint](https://prometheus.io/docs/prometheus/latest/management_api/#reload) which in turn requires starting prometheus component with `--web.enable-lifecycle` flag set.

```release-note
Update componentsPatches extraArgs Value from string to *string to distinguish nil (bare flag) from empty string (explicit --key=).
```


pkg/kwokctl/runtime/util.go
```diff
@@ -118,9 +118,12 @@ func applyComponentPatch(ctx context.Context, component *internalversion.Compone
...
 	for _, a := range patch.ExtraArgs {
-		if a.Override {
+		switch {
+		case a.Override:
 			component.Args = applyComponentArgsOverride(ctx, component.Args, a)
-		} else {
+		case a.Value == "":
+			component.Args = append(component.Args, fmt.Sprintf("--%s", a.Key))
+		default:
 			component.Args = append(component.Args, fmt.Sprintf("--%s=%s", a.Key, a.Value))
```
Priyankasaggu11929: changed if/else chain to switch

based on suggestion from failing golangci-lint CI 
https://github.com/kubernetes-sigs/kwok/actions/runs/29824210224/job/88613667202?pr=1722#step:4:34

wzshiming (commented):
`--flag=` It is meaningful in some scenarios, I think we can add a new field to indicate ignoring the value, or change the value type to `*string`, when it is null, it will be ignored.


https://github.com/search?q=org%3Akubernetes+%22-+--cluster%3D%22++&type=code

Priyankasaggu11929:
> `--flag=` It is meaningful in some scenarios, I think we can add a new field to indicate ignoring the value, or change the value type to `*string`, when it is null, it will be ignored.
> 
> https://github.com/search?q=org%3Akubernetes+%22-+--cluster%3D%22++&type=code

Thanks for the pointer.

I'll update PR to change the value type to `*string`

Priyankasaggu11929:
Hello @wzshiming, ready for another review. Thank you.

pkg/kwokctl/runtime/util.go
```diff
@@ -131,7 +134,11 @@ func applyComponentArgsOverride(ctx context.Context, args []string, a internalve
 	overrided := false
 	for i := len(args) - 1; i >= 0; i-- {
 		if strings.HasPrefix(args[i], k) {
```
wzshiming: Add checking for the original argument might be a bare boolean flag

```
	keyPrefix := fmt.Sprintf("--%s=", a.Key)
	keyFlag := fmt.Sprintf("--%s", a.Key)
```


```suggestion
		if args[i] == keyFlag || strings.HasPrefix(args[i], keyPrefix) {
```

wzshiming (commented):
There’s one more small change needed; everything else is fine.

pkg/kwokctl/runtime/util.go
Priyankasaggu11929: Thank you. Addressed in latest commit - https://github.com/kubernetes-sigs/kwok/pull/1722/commits/393ce6c56a3b648f373124e82a39b8c230083322

wzshiming (approved):
Thank you

Priyankasaggu11929:
Thank you for the review @wzshiming.

I am not able to pinpoint anything back to the PR changes that might be related to the failing job (it seems flaking across tests right now.)

re-triggering it once again

/test pull-kwok-e2e-test-main

PR #1747, opened 2026-08-31
[kwokctl] Keep binary port-forward alive after dial failure

### LINKED ISSUE #1740: kwokctl --runtime=binary: port-forward stops serving after a single dial failure
### How to use it?

- [ ] kwok
- [ ] kwokctl --runtime=docker (default runtime)
- [x] kwokctl --runtime=binary
- [ ] kwokctl --runtime=kind

### What happened?

In the binary runtime's `PortForward` accept loop, a failed dial to the target port returns from the goroutine instead of continuing:

https://github.com/kubernetes-sigs/kwok/blob/172defa550f15c9ffd4169896c32bba4770f6577/pkg/kwokctl/runtime/binary/cluster_port_forward.go#L63-L96

```go
target, err := net.Dial("tcp", fmt.Sprintf(":%d", targetPort))
if err != nil {
	_ = conn.Close()
	logger.Error("failed to connect", ...)
	return // kills the whole accept loop
}
```

A single transient dial failure (e.g. the target component is restarting) permanently terminates the accept goroutine. The listener stays open — it is only closed by the returned cancel function — so subsequent connections are queued in the accept backlog and hang instead of being served or refused. The port-forward looks alive but is dead, even after the target recovers.

A failed dial affects only that one connection (`continue`); the loop keeps accepting and recovers as soon as the target is reachable again.

1. Create a cluster with `--runtime binary` and port-forward to a component port.
2. Stop the component, make one connection attempt (dial fails, loop exits), then restart the component.
3. Further connections hang forever even though the target is healthy again.

The `listener.Accept` error path returning is correct (that means the listener was closed); only the dial error path should `continue`. Can send a PR.

### Kwok version

Found by code inspection on `main` @ 172defa5.

### OS version

N/A — platform-independent.

/kind bug

dlanov:
#### What type of PR is this?

/kind bug

The binary runtime's port-forward accept loop stopped serving after a single failed connection to the target port.

When the target component was temporarily unavailable, the failed `net.Dial` closed the client connection and returned from the entire accept loop. The forwarding listener remained open, but subsequent connections were no longer accepted or forwarded.

This change keeps the accept loop running after a target dial failure so the same port-forward can recover when the target becomes available again.

A regression test verifies that:

* the initial connection fails while the target is unavailable
* the target is then started
* a second connection through the same port-forward successfully sends and receives data

Fixes #1740

Validation completed locally:

* `go test -race -v -run TestPortForwardRecoversAfterFailedTargetDial -count=10 ./pkg/kwokctl/runtime/binary/...`
* `go test -race ./pkg/...`
* `git diff --check`

The regression test was also verified red-before-green. With the original `return`, the second connection times out because the accept loop has exited. With the change to `continue`, the same port-forward successfully recovers and forwards data.

```release-note
[kwokctl] Binary runtime port-forward now recovers after transient target connection failures.
```

```docs
NONE
```

### DIFF
--- a/pkg/kwokctl/runtime/binary/cluster_port_forward_test.go
+++ b/pkg/kwokctl/runtime/binary/cluster_port_forward_test.go
@@ -1,0 +1,139 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package binary
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net"
+	"strconv"
+	"testing"
+	"time"
+
+	"sigs.k8s.io/kwok/pkg/kwokctl/runtime"
+)
+
+// pickUnusedPort asks the OS for an ephemeral TCP port and releases it
+// immediately so the caller can reuse the port number.
+func pickUnusedPort(t *testing.T) uint32 {
+	t.Helper()
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Fatalf("failed to allocate an unused port: %v", err)
+	}
+	port := listener.Addr().(*net.TCPAddr).Port
+	if err := listener.Close(); err != nil {
+		t.Fatalf("failed to release allocated port: %v", err)
+	}
+	return uint32(port)
+}
+
+// TestPortForwardRecoversAfterFailedTargetDial is a regression test for
+// https://github.com/kubernetes-sigs/kwok/issues/1740: a single failed dial
+// to the target port used to terminate the whole accept loop, permanently
+// killing the port-forward while the listening socket stayed open.
+func TestPortForwardRecoversAfterFailedTargetDial(t *testing.T) {
+	ctx, cancelCtx := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancelCtx()
+
+	hostPort := pickUnusedPort(t)
+	targetPort := pickUnusedPort(t)
+	for targetPort == hostPort {
+		targetPort = pickUnusedPort(t)
+	}
+	// nothing listens on targetPort yet
+
+	c := &Cluster{
+		Cluster: runtime.NewCluster("test", t.TempDir()),
+	}
+
+	cancel, err := c.PortForward(ctx, "test-component", strconv.FormatUint(uint64(targetPort), 10), hostPort)
+	if err != nil {
+		t.Fatalf("PortForward returned error: %v", err)
+	}
+	defer cancel()
+
+	fwdAddr := fmt.Sprintf("127.0.0.1:%d", hostPort)
+
+	// Step 1: connect once while nothing listens on targetPort, forcing the
+	// internal dial to the target to fail.
+	conn1, err := net.DialTimeout("tcp", fwdAddr, 5*time.Second)
+	if err != nil {
+		t.Fatalf("failed to dial forwarding port: %v", err)
+	}
+	defer func() { _ = conn1.Close() }()
+
+	// Step 2: deterministically prove the failed dial has already been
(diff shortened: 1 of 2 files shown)


pkg/kwokctl/runtime/binary/cluster_port_forward_test.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	buf := make([]byte, 1)
+	if _, err := conn1.Read(buf); err == nil {
+		t.Fatalf("expected connection to be closed after failed target dial, but read succeeded")
+	}
+	_ = conn1.Close()
+
+	// Step 3: bring the target up and prove the SAME PortForward instance
+	// can still serve a new connection.
+	targetLn, err := net.Listen("tcp", fmt.Sprintf(":%d", targetPort))
```
wzshiming:
```suggestion
	targetLn, err := net.Listen("tcp", ":0")
```

Move this to the first step

wzshiming (changes requested):

pkg/kwokctl/runtime/binary/cluster_port_forward_test.go
dlanov: Updated as suggested. The target port is now allocated with net.Listen(":0") at the start, then closed before the first connection so the failed-dial regression is still exercised.

pkg/kwokctl/runtime/binary/cluster_port_forward_test.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	targetPort := uint32(targetLn.Addr().(*net.TCPAddr).Port)
+	if err := targetLn.Close(); err != nil {
+		t.Fatalf("failed to release target port: %v", err)
+	}
+
+	hostPort := pickUnusedPort(t)
+	for hostPort == targetPort {
+		hostPort = pickUnusedPort(t)
+	}
```
wzshiming: Use sigs.k8s.io/kwok/pkg/utils/net
```suggestion
	hostPort := net.GetUnusedPort(ctx, nil)
```

pkg/kwokctl/runtime/binary/cluster_port_forward_test.go
```diff
@@ -0,0 +1,151 @@
...
+	// FIRST dial to targetPort fails, reproducing #1740; targetLn is reused
+	// below once the target comes up on this same port number.
+	targetLn, err := net.Listen("tcp", ":0")
+	if err != nil {
+		t.Fatalf("failed to allocate target port: %v", err)
+	}
+	targetPort := uint32(targetLn.Addr().(*net.TCPAddr).Port)
+	if err := targetLn.Close(); err != nil {
+		t.Fatalf("failed to release target port: %v", err)
+	}
```
wzshiming, line 52: Why not just use this listener?

wzshiming (changes requested):

pkg/kwokctl/runtime/binary/cluster_port_forward_test.go
dlanov: Closing it is intentional. If this listener stays open, PortForward's first net.Dial(targetPort) can succeed against the listening socket even before Accept is called, so the failed-dial path from #1740 would not be exercised. The test closes it before the first forwarded connection, then rebinds the same port after confirming the failed dial was processed.

pkg/kwokctl/runtime/binary/cluster_port_forward_test.go
dlanov: Updated. hostPort now uses utilsnet.GetUnusedPort(ctx, sets.NewSets(targetPort)); the used-port set keeps hostPort distinct from targetPort, and the returned error is handled explicitly.

pkg/kwokctl/runtime/binary/cluster_port_forward_test.go
```diff
@@ -0,0 +1,151 @@
...
+	// to the accept loop afterward differs), so observing the close here -
+	// bounded by a deadline instead of an arbitrary sleep - is a reliable,
+	// version-independent synchronization point.
+	if err := conn1.SetReadDeadline(time.Now().Add(5 * time.Second)); err != nil {
+		t.Fatalf("failed to set read deadline: %v", err)
+	}
+	buf := make([]byte, 1)
+	if _, err := conn1.Read(buf); err == nil {
+		t.Fatalf("expected connection to be closed after failed target dial, but read succeeded")
+	}
```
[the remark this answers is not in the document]
dlanov: Updated. The test now treats a read timeout as a failure: it only proceeds after conn1.Read returns a non-timeout error, so the target listener cannot start until PortForward has actually processed and closed the first failed connection.

PR #1754, opened 2026-09-06
[kwokctl] Return scale deletion errors

immanuwell:
#### What type of PR is this?

/kind bug

`kwokctl scale` swallowed DELETE failures, logged `Deleted resources`, and exited 0 while objects remained. 
Tiny footgun

Keep processing the batch, count successful deletes, then return all API errors

Repro:

```bash
kwokctl scale node denied-node --replicas 1
kubectl apply -f - <<'EOF'
apiVersion: v1
kind: List
items:
- apiVersion: admissionregistration.k8s.io/v1
  kind: ValidatingAdmissionPolicy
  metadata:
    name: deny-node-delete
  spec:
    matchConstraints:
      resourceRules:
      - apiGroups: [""]
        apiVersions: ["v1"]
        operations: ["DELETE"]
        resources: ["nodes"]
    validations:
    - expression: "false"
      message: "node deletion denied"
- apiVersion: admissionregistration.k8s.io/v1
  kind: ValidatingAdmissionPolicyBinding
  metadata:
    name: deny-node-delete
  spec:
    policyName: deny-node-delete
    validationActions: [Deny]
EOF
kwokctl scale node denied-node --replicas 0
echo $?
```

Before this exits 0. 
With this change it exits 1 and reports the denial, as expected

None

Tested with `go test -race ./pkg/kwokctl/scale`, `go test ./pkg/...`, and `make verify`.

```release-note
`kwokctl scale` now returns an error when a resource cannot be deleted during scale down.
```

```docs
NONE
```

### DIFF
--- a/pkg/kwokctl/scale/scale.go
+++ b/pkg/kwokctl/scale/scale.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"sort"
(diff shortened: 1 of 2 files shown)


pkg/kwokctl/scale/scale.go
```diff
@@ -234,6 +222,11 @@ func Scale(ctx context.Context, clientset client.Clientset, conf Config) error {
 			"counter", deleteCount,
 			"elapsed", time.Since(start),
 		)
+	}
+	if err := errors.Join(deleteErrors...); err != nil {
+		return err
+	}
```
wzshiming, line 228: This is a behavioural change.
And according to the behaviour of kubectl scale, scale down failed will not cause response to exit incorrectly

immanuwell: @wzshiming hmm, are you suggesting that kwokctl scale should preserve the current exit-0 behavior and that this PR should be closed?

Issue #1738, opened 2026-08-26
Flaky e2e: 10s wait timeouts in jsonpatch and workable tests


wzshiming:
### Which jobs are failing?

Not continuously failing — this is a flake report. Observed on prow `pull-kwok-e2e-test-main`. The same short timeouts are also exercised by the GitHub Actions e2e jobs.

### Which tests are failing?

- `TestJsonpatch/Jsonpatch Stage/test node jsonpatch` (e2e/kwok/default) — the observed failure
- `test pod jsonpatch` and the `kwokctl` workable check share the same 10s pattern and can flake the same way

### Since when has it been failing?

Flake observed 2026-08-21 on PR #1733 at 369d9359: `pull-kwok-e2e-test-main` failed in `test node jsonpatch`, while `/retest` on the exact same commit passed with zero code changes (and the GitHub Actions e2e job on that commit passed as well).

### Reason for failure (if possible)

Three waits use a 10s timeout while equivalent waits in the same files use 600s:

- https://github.com/kubernetes-sigs/kwok/blob/172defa550f15c9ffd4169896c32bba4770f6577/test/e2e/jsonpatch.go#L145 (node phase)
- https://github.com/kubernetes-sigs/kwok/blob/172defa550f15c9ffd4169896c32bba4770f6577/test/e2e/jsonpatch.go#L179 (pod phase)
- https://github.com/kubernetes-sigs/kwok/blob/172defa550f15c9ffd4169896c32bba4770f6577/test/e2e/workable.go#L69

Under prow DinD load the controller can take more than 10s to apply the stage. In the failing run, the pod jsonpatch stage in the same run applied within seconds — the controller was busy, not stuck.

### Anything else we need to know?

Raising these three waits to the 600s used elsewhere in the same files (or at least 120s) should remove the flake without masking real regressions, since the assertions are on eventual state. Can send a PR.

/kind flake

PR #1758, opened 2026-09-11
Fix TestJsonpatch race by re-triggering Stage evaluation and not caching on LastSyncResourceVersion

akashchamp:
#### What type of PR is this?

/kind flake

`TestJsonpatch` is racy because kwok matches Stages against an object only when it receives an event for that object, and nothing re-evaluates existing objects when the set of Stages changes. The Stages created by the test can therefore become effective after the last event of their target has already been processed, and the target is never patched:

- `stage-jsonpatch-available` is both a rule and its own target. If the Stage-kind controller evaluates it as a target before it is visible as a rule, it matches nothing and is dropped.
- Even once that annotation has flipped, it only proves the first Stage was loaded. If `node-jsonpatch-available` is not in the Node controller's lifecycle when the node's post-initialize event is evaluated, the node falls into `node-heartbeat-with-lease` (600s delay) instead; the pod matches nothing and is never revisited.

Raising the timeouts does not address this, so the earlier timeout bumps are dropped.

Test side: every wait in `test/e2e/jsonpatch.go` now touches its target (rewrites an annotation no Stage selects on) between polls, so kwok receives a new event for the object and matches it again against the Stages it currently has. The waits are bounded at two minutes.

kwok side: `dynamicGetter.Version()` no longer returns the informer's `LastSyncResourceVersion`. The reflector advances that as soon as an event is queued, before the object reaches the store, so a `Get` in that window cached a snapshot without the new Stage under the new version and kept serving it until the next Stage event (or bookmark). `Version()` is now a counter bumped from the informer event handlers, which run after the store has been updated. `TestDynamicGetterVersionFollowsStore` covers this.

The `workable.go` change from the first version of this PR is reverted: that check is a different code path (stages come from the kwokctl config) and is not covered by this root cause.

Related to #1738

Re-evaluating managed objects when the Stage set changes (direction (b) in the issue) is not attempted here; the test no longer depends on it, and the cache fix removes the tens-of-seconds staleness that made the window wide.

#### How was this tested?

- `go build ./...`, `go vet ./pkg/... ./test/e2e/...`
- `go test ./pkg/config/resources/...` (new unit test)
- `go test -c` for `./test/e2e/kwok/default` and `./test/e2e/kwokctl/binary`
- `hack/verify-boilerplate.sh`, `hack/verify-ends-newline.sh`, `hack/verify-go-format.sh` (also with Go 1.27.1 on a tree merged with `main`), `hack/verify-spelling.sh`
- The cluster e2e (`TestJsonpatch` on kind, and the kwokctl `Workable` check) was not run locally for this revision; docker and kind are not available on the machine used, so it relies on `pull-kwok-e2e-test-main`. The first revision of this PR had been run manually against a kind cluster, but that revision has been replaced.

```release-note
NONE
```

```docs
NONE
```


wzshiming:
Sorry, this issue was created by AI, which got the root cause wrong, and the issue has now been updated.

akashchamp:
Thanks for the corrected analysis, I have reworked the PR accordingly and dropped the timeout bumps and the workable.go change. In test/e2e/jsonpatch.go every wait now touches its target (a merge patch on an annotation no Stage selects on) between polls, so kwok receives a new event and matches the object again against the Stages it currently has, with each wait bounded at two minutes. On the kwok side, dynamicGetter.Version() is now a counter bumped from the informer event handlers instead of LastSyncResourceVersion, so the cached lifecycle can no longer latch a snapshot that is missing a Stage the store already has; there is a unit test for that in pkg/config/resources/dynamic_test.go. I left re-evaluating managed objects when the Stage set changes (direction b) as a possible follow-up.

wzshiming (changes requested):
This should only modify the test cases

akashchamp:
Done: dropped the `DynamicGetter.Version()` fix and its test, this PR now only touches `test/e2e/jsonpatch.go`. The touch-and-retry wait already makes the barrier correct by itself: every poll re-touches the target so a Stage that only became visible after the target's last event still gets picked up on the next retry, independent of how exact `Version()` is. The version fix was a real bug I found while chasing this (kwok never keys the cache invalidation off the store write itself), but it's not needed here, so I've left it out.

